Issue #57, opened 2016-02-02
pandoc-citeproc: could not find locale data for american


mvuorre:
Hi,
When compiling a document (just the provided template), I get the following error in the R Markdown console:

```
pandoc-citeproc: could not find locale data for american
pandoc: Error running filter /usr/local/bin/pandoc-citeproc
Filter returned error status 1
Error: pandoc document conversion failed with error 83
Execution halted
```
However, changing `lang: american` to `lang: english` in template YAML row 31 removes the problem. If this behavior happens for others (perhaps after pandoc update), you might consider changing the default to `lang: english`.

I'm running pandoc 1.16.0.2 (updated to latest a few days ago). Session info below:

```r
> sessionInfo()
R version 3.2.2 (2015-08-14)
Platform: x86_64-apple-darwin13.4.0 (64-bit)
Running under: OS X 10.11.3 (El Capitan)

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] rsconnect_0.4.1.11 htmltools_0.3      tools_3.2.2        yaml_2.1.13        rmarkdown_0.9.2   
[6] papaja_0.1.0.9054  digest_0.6.9  
```

Issue #56, opened 2016-01-29
Problems with brewed pandoc


sebastiansauer:
Hi there,

I have tried to knit the `example.rmd` file using RStudio (0.99.491), pandoc (1.16.0.2) with pandoc-citeproc (0.9). This resulted in the following error:

     pandoc-citeproc: /usr/local/Cellar/pandoc-citeproc/0.9/share/x86_64-osx-ghc-7.10.3/pandoc-citeproc-0.9/locales/locales-am.xml: openBinaryFile: does not exist (No such file or directory)
pandoc: Error running filter /usr/local/bin/pandoc-citeproc

Any ideas what the problem could be?

Thanks for any advise!

Best
Sebastian

crsh:
Thanks for reporting. I don't have time to follow up right now, but I will take a closer look on Monday. I suspect it may have something to do with a change that was made to locale identifiers in one of the recent pandoc updates.

sebastiansauer:
Hi Frederik,

no worries, and let me add that I think your APA paper package is of potentially great help. I hope I get it running. Also, I am a markdown/pandoc fan.

First, I thought it may be due to Homebrew. But even without it, it did not work. I think your point makes sense.

You could also try to use the package „checkpoint“ to have reproducibility as to package versions.

Best
Sebastian

Beste Grüße/ Kind Regards

Sebastian Sauer

crsh:
If you haven't set `lang`, try setting it to `de-DE` or `en-US` and see if that works. Could you also post the output of `sessionInfo()` and your YAML header?

sebastiansauer:
Hi Frederik,
thanks for your help. I am on a business trip, and had no time ^for looking into it. Will get 	back  to your points asap!
Thanks!
Sebastian

Beste Grüße/ Kind Regards

Sebastian Sauer

crsh:
No problem. In the mean time @mvuorre reported a similar problem and was so kind to follow up with a pull request to fix the problem. Try changing `lang` in the YAML header to `english`. I think this should take care of the problem. Let me know if the problem persists.

mvuorre:
Also might be best to remove or comment out the `lang: ` line as it probably doesn't do much anyway.

sebastiansauer:
Hi Matti, hi Frederik,

thanks! Removing „lang:“ out from the YAML header has solved the problem. It now compiled!

Thanks!

Great work! Please keep up!
Sebastian



> sessionInfo()
[...]

PR #58, opened 2016-02-02
Update skeleton.Rmd

mvuorre:
`lang: american` to `lang: english` in YAML

This fixes issue #57 (if persists for other users after updating pandoc)


mvuorre:
issue #57

crsh:
Thanks for following up on this one. I suspect this should also fix #56.

Issue #108, opened 2017-04-08
Omit author note


jacob-long:
As far as I can tell, even if I omit author note in the YAML frontmatter, there is still an "Author Note" heading followed by no content on the title page. I know the APA6 class won't print it if it isn't there in the LaTeX, but it apparently is included in the LaTeX by `papaja` even when it's omitted in the frontmatter.

jacob-long:
Okay, poking around in your `apa6.tex` file, I can see what's going on on the TeX side. You do detect if there's no author note in the frontmatter, but fall back on printing information about the corresponding author in that case. I set the sole author in my frontmatter to not be the corresponding author, so this part of the conditional just generates empty content:

    $else$ % If no author_note is defined give only author information if available
      $if(author)$
      \authornote{
        \newcounter{author}
        $if(author)$
        $for(author)$
          $if(author.corresponding)$
            Correspondence concerning this article should be addressed to $author.name$$if(author.address)$, $author.address$$endif$.$if(author.email)$ E-mail: $author.email$$endif$
         $endif$
        $endfor$
        $endif$
      }
      $endif$
    $endif$

Maybe you could check for whether any listed author is corresponding first, then call `\authornote`.

crsh:
Thanks for the pull request, I've merged your changes. Does this resolve this issue?

jacob-long:
Yes!

TtecSiyu:
When I set the sole author to not be the corresponding author, I got missing Missing \begin{document} error

crsh:
This should be fixed in the current development version (`devtools::install_github("crsh/papaja@devel")`).

PR #109, opened 2017-04-08
Don't use author note if there's no corresponding author.

jacob-long:
This addresses #108. I tested it with a few author configurations and it seems to preserve functionality otherwise.


crsh:
Hi Jacob, thanks for the pull request. Since APA style requires a corresponding author I had not explored this edge case. I guess if a user *really* doesn't want the required line or would rather generate it themselves this should do it.

jacob-long:
Just to give you a sense of why I was trying to have that flexibility in the first place: I am a graduate student and in the universities I've been affiliated with (all in the USA, FWIW) it is not necessarily prohibited but is certainly unusual to give contact information for a corresponding author in a student paper.

Issue #116, opened 2017-05-25
apa_table() breaks if column names contain underscores


mvuorre:
Hi,

`apa_table(d)` breaks when `names(d)` contains underscores.

You can reproduce this problem by creating a new papaja template document and trying to knit it to PDF with this content:

````
```{r}
d <- data.frame(v1 = 1, v_2 = 1, v_3 = 1, v4 = 1)
```

```{r, results = "asis"}
# This breaks
apa_table(d)
```
````

However, if you get rid of the underscores it works fine. For example, you could do the following for every data.frame:

````
```{r, results = "asis"}
# This works
names(d) <- gsub("_", "\\_", names(d), fixed = TRUE)
apa_table(d)
```
````

mvuorre:
Pull request #116 has a quick fix for this.

mvuorre:
The same issue holds for certain special symbols (percentages and underscores) anywhere in the table that's input to `apa_table.latex()`. I've updated the pull request to deal with those as well, but it would be good to see if there is a general solution instead.

crsh:
Hi Matti, have you tried setting `escape = TRUE`?

mvuorre:
Wow, I had no idea kable had that as an option (and being lazy did not go beyond the apa_table() help page.) That works perfectly, thanks!

(Is there a way to remove the "Note." from the bottom of the tables?)

PR #117, opened 2017-05-25
Fix underscores in column names for apa_table.latex()

mvuorre:
This fixes issue #116


mvuorre:
The fix needs to be generalized to row names and other special symbols.

PR #153, opened 2017-10-27
Eta squared

mariusbarth:
The calculation of effect-size measures and mean squared errors for ANOVA objects now happens in a specialized function `add_effect_sizes()` that can be found in `calculations.R`. I also implemented a first test for a two-way between-subjects design. Inclusion or exclusion of the intercept is intended to produce different results for eta-squared, but not generalized or partial eta-squared. All already existing tests for ANOVA should pass.


R/arrange_anova.R
```diff
@@ -90,13 +90,19 @@ arrange_anova.summary.aov <- function(x) {
 
   # When processing aovlist via lapply a dummy term "aovlist_residuals" preserves the SS_error of the intercept
   # term to calculate generalized eta squared correctly later on.
+  # ----
+  # We now changed this to the term (Intercept) -- for the sake of consistency
+  # and standardized processing later on
+
+  # When processing an aovlist, this one-row aov-object contains the sum of squares
```
crsh: The residual sum of squares, right?

R/calculations.R
```diff
@@ -256,3 +256,93 @@ hd_int <- function(x, level = 0.95) {
...
+#'
+#' @param x An object of class \code{apa_variance_table}.
+#' @param es Character. A vector naming all to-be-computed effect-size measures.
+#'   Currently, partial eta-quared (\code{"pes"}), generalized eta-squared
+#'   (\code{"ges"}), and eta-squared (\code{"es"}) are supported.
+#' @param observed Character. A vector naming all factors that are observed
+#'   (i.e., \emph{not} manipulated).
+#' @param mse Logical. Should means-squared errors be computed?
+#' @param intercept Logical. Should the sum of squares of the intercept (i.e., the
+#'   deviation of the grand mean from 0) be included in the calculation of eta-squared?
```
crsh: Maybe we should add something like "Set to true if a deviation of the grand mean from zero constitutes to-be-explained variance."?

R/calculations.R
```diff
@@ -256,3 +256,93 @@ hd_int <- function(x, level = 0.95) {
...
+
+add_effect_sizes <- function(x, es = "ges", observed = NULL, mse = TRUE, intercept) {
```
crsh: I think a default of `intercept = FALSE` would make sense here.

R/calculations.R
```diff
@@ -256,3 +256,93 @@ hd_int <- function(x, level = 0.95) {
...
+    # individual observations from the grand mean (the general specification)
+    # would be highly desirable. However, most ANOVA outputs do not provide
+    # the necessary information, so we have to go with this hack.
+    if("es" %in% es) {
+      index <- rep(TRUE, nrow(x))
+      if(!intercept){
+        index <- x$term!="(Intercept)"
+      }
+      x$es <- x$sumsq / sum(x$sumsq[index], unique(x$sumsq_err))
+      message("For your information: Eta-squared is calculated correctly if and only if the design is balanced.")
```
crsh: I'd opt for "Note that ..."

R/print_anova.R
```diff
@@ -7,6 +7,7 @@
 #' @param x Data.frame. A \code{data.frame} of class \code{apa_variance_table} as returned by \code{\link{arrange_anova}}.
 #' @param intercept Logical. Indicates if intercept test should be included in output.
 #' @param es Character. The effect-size measure to be calculated; can be either \code{ges} for generalized eta-squared, \code{pes} for partial eta-squared or \code{es} for eta-squared.
+#'   Be careful when calculating eta-squared: It is only calculated correctly if the design is balanced.
```
crsh: Could we reiterate the contents of the message given by the function here?

tests/testthat/test_apa_print_anova.R
```diff
@@ -243,21 +243,21 @@ test_that(
...
     expect_identical(rm_afex_aov_output, rm_aov_output)
 
-    rm_afex_aov_output <- apa_print(rm_afex_aov, correction = "none")
+    rm_afex_aov_output <- apa_print(rm_afex_aov, correction = "none", intercept = FALSE)
     expect_identical(rm_afex_aov_output, rm_aov_output)
 
 
     # DF corrections
-    rm_afex_anova.mlm_output <- apa_print(rm_afex_aov$Anova, correction = "GG")
+    rm_afex_anova.mlm_output <- apa_print(rm_afex_aov$Anova, correction = "GG", intercept = TRUE)
```
crsh: Do we need to set `intercept = TRUE` in these tests? Shouldn't make any difference, no?

R/arrange_anova.R
mariusbarth: Yep, you're right.

R/calculations.R
mariusbarth: Will do that.

R/calculations.R
mariusbarth: Yep, this seems to be more consistent with what we usually read.

R/print_anova.R
mariusbarth: I like that idea.

R/calculations.R
mariusbarth: Hmm, dunno. I think this sounds a bit too much like a well-justified advice in statistical reasoning, something I don't claim for my within-package documentation.

tests/testthat/test_apa_print_anova.R
mariusbarth: Totally true, removed these changes.

R/calculations.R
crsh: Fair enough. :)

crsh:
I'm merging this. Again, thanks for the work you put into this and the quick response.

PR #169, opened 2017-12-26
`apa_sinaplot`

mariusbarth:
This year's Christmas, our users get a new plot type `apa_sinaplot()`, inspired by the `sinaplot` package. Moreover, our aggregation function now supports arguments that can be passed to the aggregation function (e.g., the `level` argument to `conf_int`). This also simplifies the code in `apa_factorial_plot`.

(+ 1 fixed typo)


R/apa_factorial_plot.R
```diff
@@ -277,11 +281,15 @@ apa_factorial_plot.default <- function(
   if(fun_dispersion == "within_subjects_conf_int" || fun_dispersion == "wsci") {
     ee <- wsci(data = aggregated, id = id, factors = factors, level = level, method = "Morey", dv = dv)
   } else {
-    if(fun_dispersion == "conf_int") {
-      ee <- stats::aggregate(formula = stats::as.formula(paste0(dv, "~", paste(factors, collapse = "*"))), data = aggregated, FUN = dispersion, level = level)
-    } else {
-      if(use_dplyr) {
+    if(use_dplyr) {
+      if(fun_dispersion=="conf_int"){
+        ee <- fast_aggregate(data = aggregated, factors = factors, dv = dv, fun = dispersion, level = level)
```
crsh: Could we make this more generic so that we add an option to pass arguments to the dispersion function (e.g., `args_dispersion`)? That would make it possible to choose a method for `wsci()`, for example. Another application  I can think off would be to pass a dispersion function that computes bootstrap confidence intervals and would take options such as the number of bootstrap samples.

crsh:
I will close this PR for now. I'd really like to see sina plots in `papaja`. I think they are the mix of bee and density plots. So I'd be more than happy to review another pull request that addresses the above issues.

PR #185, opened 2018-02-22
Variable labels

mariusbarth:
Hi @crsh,
I think everything is ready for switching to S3 variable labels.


tests/testthat/test_variable_label.R
```diff
@@ -140,19 +52,96 @@ test_that(
...
+    variable_label(y) <- "Test me!"
+
+    expect_identical(
+      object = x
+      , expected = y
+    )
+  }
+)
+
+test_that(
```
crsh: For some reason, this test fails for me when I run it manually. The indexing method also does not work in the console for me. Any idea what might me happening?

mariusbarth: Did you mean the "`[`.labelled-method" test? This one should work now

R/variable_label.R
```diff
@@ -1,267 +1,139 @@
...
-  , definition = function(object, value){
-    object@annotation@label <- value
-    validObject(object@annotation)
-    object@label <- value
-    validObject(object)
-    object
-  }
-)
+`variable_label<-.data.frame` <- function(x, value){
+  assign_label.data.frame(x, value)
```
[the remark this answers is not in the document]
mariusbarth: This should work by the method dispatch of `assign_label()`. I changed line 104 accordingly.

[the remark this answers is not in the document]
mariusbarth: Naaa, it seems like we have to specify assign_label.default() if called from within assign_label.data.frame()... Could this become problematic?

tests/testthat/test_variable_label.R
crsh: Yes, that's what I meant.

crsh:
Great, thanks. Looks good now. Before I merge: Any idea why the tests past previously although there was not `[.labelled` method? This seems to indicate that the tests are not diagnostic...

mariusbarth:
tbh, dunno

crsh:
Hmm, I feel like it is a bad idea to have a test that is non-diagnostic. It gives a false sense of security. So I think we should either remove the test entirely or (preferably) make sure that it fails as expected.

To clarify, the [test](https://github.com/crsh/papaja/pull/185/files/129f9161641a9014427a87a4af59018c32145db0#diff-117bfe4772ffd8ba213b474cbc9b627eR89) of the `[` method passed when I tested the whole package although the method did not exist. It only failed when I ran the tests manually by executing them line by line.

mariusbarth:
I fiddled around with the non-diagnostic test, but I'm running out of ideas how to de-activate the `[.labelled`-method from another package (probably `Hmisc`). I'm a bit annoyed by this, and also wondering if we should maybe change the class name to `"annotated"` to avoid any future conflicts with other packages.

crsh:
Okay, that was tricky. But I was able to address this by moving the tests forward. I have had to load the `BayesFactor` package during the tests, which somehow resulted in the registration of the `Hmisc` S3 methods for the `labelled` class. Now that the order of the tests has changed, they should be diagnostic.

Thanks again!

PR #197, opened 2018-03-28
Fix link

mvuorre:
Previous link kept taking me to https://github.com/crsh/papaja/blob/master/github.com/crsh/citr, which is a 404.


crsh:
Thanks Matti!

Issue #201, opened 2018-04-11
apa_print.summary.Anova.mlm() fails


mariusbarth:
When I do this
```
anova_out <- aov_ez(data = npk, id = "block", dv = "yield", within = "N", return = "univariate")
apa_print(anova_out)
```
I get the follwing error:
```
Error in `[.data.frame`(variance_table, , univariate_names) : undefined columns selected
```

Probably Henrik changed the column names of his output object in `afex`. I will try to fix this, asap.

mariusbarth:
I just committed that bugfix. However, when running unit tests, I also figured out that the tables from `test_apa_table.tex` look weird (at least unfamiliar). @crsh, could you also get weird tables (maybe it's my current tex config)?

PR #202, opened 2018-04-11
Bugfix, solves #201

mariusbarth:
- changed the renaming of columns in apa_print.summary.Anova.mlm to a more flexible (broom-inspired) coding
- added support for new `afex_ov` objects to `apa_factorial_plot()`
- changed the source of the Moore dataset in tests (as it moved from `car` to `carData`)
- added `_config.yml` to `.Rbuildignore`


crsh:
Thanks!

Issue #39, opened 2015-06-15
Confidence intervals for eta^2s


crsh:
`apa_print()` provides confidence intervals for R² (`lm`) but as of now no confidence intervals are calculated for ANOVA objects.

crsh:
http://daniellakens.blogspot.de/2014/06/calculating-confidence-intervals-for.html

crsh:
Also see this code by Richard Morey: https://gist.github.com/richarddmorey/ea32db6a6c2419eaff0e1a5186e986c0

doomlab:
Our package has this functionality now, and I am adding omega and epsilon soon. 
I suppose the biggest question is if you want to just get people to use the MOTE or MBESS package to calculate them and then make sure their output is convertable using apa_print or add the calculation functionality directly to the papaja package.

crsh:
So far, my idea was to give users the opportunity to specify a function to calculate effect sizes and confidence intervals as an argument to `apa_print()`. I'm not sure if this approach is feasible, though. Is [this](https://github.com/doomlab/MOTE/blob/2265c1bdef59a558ce33d6fb3eb049cf123def4c/R/eta.anova.R) the function in `MOTE` that does this?

doomlab:
@crsh - sorry for the delay - yes that function will do a basic eta translation. One of my goals over the next couple of weeks is writing more functions for MOTE that will cover the gambit of eta/omega/epsilon/ges, so there will be more options than just calculating from df and F. 

Is there a good way to write those functions to pass to `apa_print()` ... this is probably where I need to read and learn the `S3` stuff a bit better yes? Trying to do double duty here on finishing some work that will help me learn/understand how to build better functions that I can apply to helping with your package.

crsh:
Hi @doomlab,

generally, it would be helpful to define an S3 or S4 class for the output objects. It makes sense to invest some thought into whether there is some structure that works for all related functions so that the results can be returned in a standardized form. This will facilitate the processing of the results in `apa_print()`. For example, a useful default structure would be a [tidy `data.frame`](https://cran.r-project.org/web/packages/tidyr/vignettes/tidy-data.html). I'd take a look at the objects returned by [`broom`-functions](https://github.com/tidyverse/broom/blob/master/R/anova_tidiers.R) as examples for how to name the columns.

[...]

PR #212, opened 2018-05-12
[WIP] Confidence intervals to ANOVA effect sizes

doomlab:
Added confidence intervals to the es/pes/ges calculations for aov like objects.

### DIFF
--- a/R/calculations.R
+++ b/R/calculations.R
@@ -280,20 +280,38 @@ hd_int <- function(x, level = 0.95) {
 #' @param intercept Logical. Should the sum of squares of the intercept (i.e., the
 #'   deviation of the grand mean from 0) be included in the calculation of eta-squared?
 
-add_effect_sizes <- function(x, es = "ges", observed = NULL, mse = TRUE, intercept = FALSE) {
+add_effect_sizes <- function(x, es = "ges", observed = NULL, mse = TRUE, intercept = FALSE, ci_conf_level = 95) {
   # ----------------------------------------------------------------------------
   # We don't validate here because this function is intended to be used
   # internally, validation, should have happened earlier in the processing chain.
-
+  
   # validate(x, check_class = "apa_variance_table", check_NA = FALSE)
   # validate(es, check_class = "character", check_NA = FALSE)
-
+  
+  # --------------------------------------------------------------------------
+  # Calculate confidence interval function 
+  
+  calc_ci <- function (x, es) {
+    
+    es_ci <- MBESS::ci.R2(
+      R2 = x[es],
+      df.1 = x["df"],
+      df.2 = x["df_res"],
+      conf.level = ci_conf_level / 100
+    )
+    
+    if(!is.na(es_ci$Lower)){ LL <- es_ci$Lower } else { LL <- 0 }
+    if(!is.na(es_ci$Upper)){ UL <- es_ci$Upper } else { UL <- 1 }
+    
+    return(c(LL, UL))
+  }
+  
   if(!is.null(es)) {
     # Stop if the user requires a non-supported effect-size measure ----
     if(!all(es %in% c("pes", "ges", "es"))) {
       stop("Requested effect size measure(s) currently not supported: ", paste(es, collapse = ", "), ".")
     }
-
+    
     # --------------------------------------------------------------------------
     # Calculate generalized eta-squared
     #
@@ -316,8 +334,12 @@ add_effect_sizes <- function(x, es = "ges", observed = NULL, mse = TRUE, interce
         obs_SSn2 <- 0
       }
       x$ges <- x$sumsq / (x$sumsq + sum(unique(x$sumsq_err)) + obs_SSn1 - obs_SSn2)
-    }
 
+      x[ , c("es_ll", "es_ul")] <- t(apply(x[,c("ges","df", "df_res")], 1, 
+                                           function(y) calc_ci(y, es = "ges")))
+
+    }
+    
     # --------------------------------------------------------------------------
     # Calculate eta-squared
     #
@@ -334,23 +356,31 @@ add_effect_sizes <- function(x, es = "ges", observed = NULL, mse = TRUE, interce
       }
       x$es <- x$sumsq / sum(x$sumsq[index], unique(x$sumsq_err))
       message("Note that eta-squared is calculated correctly if and only if the design is balanced.")
+      
+      x[ , c("es_ll", "es_ul")] <- t(apply(x[,c("es","df", "df_res")], 1, 
+                                           function(y) calc_ci(y, es = "es")))
+      
     }
-
+    
     # --------------------------------------------------------------------------
     # Calculate partial eta-squared
     #
     # This one should be unproblematic and work in all cases.
     if("pes" %in% es) {
       x$pes <- x$sumsq / (x$sumsq + x$sumsq_err)
+      
+      x[ , c("es_ll", "es_ul")] <- t(apply(x[,c("pes","df", "df_res")], 1, 
+                                           function(y) calc_ci(y, es = "pes")))
+      
     }
   }
-
+  
   # ----------------------------------------------------------------------------
   # Only calculate MSE if required (otherwise, Levene tests give an error).
   if(mse) {
     x$mse <- x$sumsq_err / x$df_res
   }
-
+  
   x
 }
 


doomlab:
@crsh - think I finally figured out how to add these! You can use the same basic ideas from R2 in MBESS, as you were already doing. I had to add ci_conf_level, as I wasn't sure if it was already encoded somewhere else (doesn't appear to be). 

I know this will also effect the code downstream in print_anova to print them out. If you liked how I coded this, I could work on adding the print out there.

crsh:
Hi  Erin,

great thanks. Is this approach viable for within-subject and mixed designs? I remember reading in a (Geman) text book that for within-subject designs the calculation of the confidence intervals should take the intraclass correlation into account (p. 491; Eid, Gollwitzer, & Schmitt, 2010):

codecogseqn

Is this already the case here? I'm asking because the documentation describes regression usecases (without random effects components, I assume).

Another issue is that we should try to make the CI consistent with the F test. In case the user has requested, for example, GG-correction for nonsphericity this should be reflected in the calculation of the confidence interval (p. 479; Eid, Gollwitzer, & Schmitt, 2015):

codecogseqn 1

So, to get at a general solution here, I think we need to proceed by first calculating the appropriate noncentrality parameter. Does that make sense?

doomlab:
@crsh 

The short answer is no - which is why I started with AOV only, as that analysis in R can/should really only handle between subjects components. 

The longer answer is that it doesn't matter so much. I wish I had this paper written so I could send it to you but we are writing it right now (here's the poster we did though https://osf.io/bgbjh/). We did some simulations of between/repeated measures anovas (one and two way) across correlations, sample size, effect size (small medium large), and the number of levels of the interested effect. We then looked at full eta, partial eta, full omega, partial omega, and ges to think about the distribution of these effect sizes, the best way to estimate confidence intervals, and sampling variance. 

We did find that the non-central F distribution was the best way to estimate the confidence interval if you didn't want to simulate your data. That being said, we were originally using a noncentral F type focus that used F in the calculation like this (Smithson CI Sage Book):

```
ncpboth <- conf.limits.ncf(Fvalue, df.1 = dfm, df.2 = dfe, conf.level = (1 - a)) ##from MBESS
glow <- ncpboth$Lower.Limit / (ncpboth$Lower.Limit + dfm + dfe + 1)
ghigh <- ncpboth$Upper.Limit / (ncpboth$Upper.Limit + dfm + dfe + 1)
```
There's also the `ci.pvaf` function in `MBESS`, but it does only between/fixed designs for what looks like full eta. The biggest problem to these approaches is that there's clearly a direct translation between F and some of the statistics (eta, partial eta, some omega). Once you get into partial omega and ges, the translation is not so direct. So, it's difficult to use the F statistic from ANOVA because it might only account for the MSeffect and MSerror, while the formula for the effect size also includes other components. That's especially true for repeated measures components when partial omega or ges are controlling for the subject variance, that is not included anywhere in the F statistic. Even if a user selects the correction for repeated measures, that isn't translated anywhere into the effect size (which is another issue entirely that I think the ges was trying to solve because partial eta is incredibly upwardly biased for repeated measures and ges is almost overkill - this is another paper we are writing). 

When we tried to use F based measures - i.e. calculate the confidence interval by finding the noncentrality from the F statistic and translating back - it failed miserably. Often the effect size was not even in the range of the CI you got back. However, if you use the `ci.R2` function (as they are all in the variance overlap family), it first translates R2 into an F value, then calculates the NCP based on that F, then back to R2 using the `NCP / (NCP + df model  + df error + 1)`. This route estimates much closer, as it forces the F value to represent the effect size and everything that is controlled for in that effect size. This route appears to work well for all the effect sizes across all our simulations. So, that's why I say it doesn't matter so much - the correction for repeated measures "subjectness" or correlation appears in the effect size formula, which then is transferred to the F and back. 

Let me know if that doesn't make sense or you have other thoughts.

crsh:
Thanks for the extensive response! :astonished:

If I understand correctly, the approximation of the "proper" CI by means for R² > F > NCP > F > R² produces reasonable results even for within-subject designs. This may be an acceptable approach then.

I'm wondering, though, we are not limited to using either the effect size estimate or the F value. We have access to the sums of squares and df. Would that help to get closer to a nominal confidence interval? I'd be happy to help with this, if it's possible and seems sensible to you. Relatedly, can we make an intermediate stop in the above conversions to adjust the NCP for nonsphericity as mentioned above?

doomlab:
@crsh 

I think that the usefulness of the df/ss is in calculating the effect size for people - which you've already got coded (at least for the anova I was looking at). Not sure I'd know the best way to approximate F for the special instances of effect size that is any better than the formula MBESS is doing: `(R2/df.1)/((1 - R2)/df.2)`. We are running into the same issue with dependent t-test confidence intervals for non-dz - and the answer is a bit more obvious because often it's just a small transform of t instead of the traditional t formula:

codecogseqn
codecogseqn 1

Now, you've got a good point on multiplying the NCP by the epsilon ... that's something I can think / test to see if it better approximates for repeated measures designs but I don't have a good answer for that right now. I don't remember the CI being any better or worse for repeated measures versus between subjects based on this procedure. The biggest take away from our work was that variance overlap statistics seem to follow a beta/gamma/normal distribution, but the best estimator for CI was the noncentral F (and for the second paper it was oh boy is this nonsense biased).

doomlab:
@crsh 
I wanted to let you know I started a (hopefully) quick set of simulations for repeated measures to see if multiplying by epsilon would approximate the CI better than just the R2 version I am doing. I'll let you know the results.

doomlab:
@crsh - same results with heterogeneous data! Mostly, the F transform works better - the average difference was .01 closer for ges and .01 pes LL .04 for pes UL.

crsh:
That's really interesting! A couple of follow-up questions, if I may:

1. I'm not sure I fully understand what it means for a CI to be "better" here. Do you evaluate whether the CI includes the true effect size in 95% of cases or whether the CI-based decision (reject or retain H0, Type-1 and Type-2 errors) corresponds to the F-test (maybe this is even the same thing here, I'm not sure). Naively I'd assume that if the F-test has to be corrected to Type-1 error inflation so should the CI? I'm assuming here it's desirable to have a CI that corresponds to the corresponding NHST result.
1. If I understand correctly, the simulation deals with a one-factorial design. Did you simulate 3-6 levels for this factor as on the poster? As evident from the lower bound estimate of Epsilon the maximal degree of sphericity violation is a function of the number of levels *k* of the factor in question (1 / (*k* - 1)).
1. Did you use Greenhouse-Geißer or Huynh-Feldt estimate of Epsilon? There is the recommendation to use either GG or HF depending on the extend of the sphericitiy violation (GG if Epsilon < .75; Girden, 1992). I'm not sure how important this could be given the parameters of the simulation.

Girden, E. R. (1992). ANOVA: Repeated measures. Sage university papers. Quantitative applications in the social sciences: no. 07-084. Newbury Park, Calif.: Sage Publications.

(I'd be happy to join either paper---whatever you think would be more appropriate.)

doomlab:
@crsh - sure thing - prepare for a long answer. :) I had to rerun a couple times because math is hard some days and I forgot that. When you actually do the formula correctly "full" eta = ges for one-way repeated measures design. Whoops. 

First - on the papers, I say this conversation has really inspired a lot for our eta simulation paper, so I would like to give you credit for that. I think we can stick this on the end without too much trouble but if it gets split into two, you should still get something for both. I can add you to the repo if you want to check it out, but right now it's a bit of a cobbled mess that I need to work on writing, updating with the ideas here, and cleaning up. 

First, I calculated the CI based on:
1) the corrected F translation we've been discussing
2) the real "found" F found in the study
3) the corrected F translation with epsilon --> NCP --> NCP * gg/hf / (NCP *gg/hf + dfm + dfr + 1)

Then I looked at the percent of time that the found effect size falls within each of these CIs. I did this for the complete simulated set first (it's 108 combinations of N, correlation, levels, and effect size across 1000 sims each, so here's the basic average, but the code has the averages for each combination which can be used for the paper):

ges:
- corrected F: 0.9999537
- regular F: 0.8361574
- GG across all: 0.9496204
- HF across all: 0.967
- GG when picked: 0.4683485 
- HF when picked: 0.9958977 

pes: 
- corrected F: 0.9999537
- real F: 0.9999537
- GG across all: 0.919537
- HF across all: 0.9427222
- GG when picked: 0.279374
- HF when picked:  0.9799267

Clearly an advantage for the corrected formula for ges, while for pes corrected = found F because the formula for pes is a straight transform for found F. No real advantage for HF or GG. 

The number of times the CI contains the TRUE effect size. I feel like this is tricky because of known bias in these bad boys, but I calculated that as well (using Okada 2013 bias paper to get the true effect size, which we also discuss in our bias paper):

- ges corrected F: 0.9763981
- ges real F: 0.7766204
- ges GG when picked: 0.4336604
- ges HF when picked: 0.8858861

- pes corrected F: 0.7766204
- pes real F: 0.7766204
- pes GG when picked: 0.3457136
- pes HF when picked: 0.7393919

ges looks good for corrected F, while again, partial eta doesn't matter. 

As to your matching hypothesis testing ... I think that kind of thing works well for d (i.e., is zero in the confidence interval), but maybe not so well here given known effect size upwards bias, and the fact that the CIs are not estimated below 0. For regular and corrected F (since the multiplication by GG HF doesn't appear to work), I examined times that: p < .05 and CI_LL > 0, since I did 95% CI. 

- ges corrected: 0.7333889
- ges real F: 0.83944444
- pes either: 0.83944444 

Likely this has an interesting relationship with the small effects and large N where the CI might include 0 but the F is significant. I don't think it's a disadvantage here though, because the CI is likely telling you that the sig F is not so interesting (i.e. all the N in the world doesn't make .001 eta a good number).

Now, I can also create a summary of each possible simulated combo by averaging down the dataset. This data might represent the average of the distribution of each combination of simulation (given that the average would be the most likely thing to draw from these simulated "populations"). Then I can pretend that the 2.5% and 97.5% quartile from the 1000 simulations is the real CI I am trying to hit, as if I've simulated them from my dataset. Given that, how close are each of the estimators to the CI? I see this has kind of a comparison of the NCP F estimator in either form versus a simulated/bootstrapped CI someone might make.

For ges, the lower limit is about half-half maybe a slight advantage to the found F, while the upper limit more closely matches corrected F. For pes, these are the same CI, so doesn't matter. 

Differences in simulated CI and F estimated CI:
corrected ges LL: -0.02215811 
corrected ges UL: 0.01366626
found ges LL: 0.02991951
found ges UL: 0.08639057
pes LL: -0.01633165
pes UL:  -0.003300364

For corrected ges, the CI is a bit wider than a simulated one, while the found F ges is shifted upwards on both ends. For pes, it's bit lower on the bottom, and pretty close on the top. I imagine if I upped the number of simulations, ges would be closer but clearly there's a winner here for corrected F. 

WHEW. PS the german double S (ß) made my day. 🥇

crsh:
Hi @doomlab, it's been a while since we last discussed this. I really appreciate all the work you put into this! As we discussed in https://github.com/crsh/papaja/issues/39, I had envisioned to implement a more general approach at a later point. After some more [prototyping](https://github.com/crsh/papaja/issues/423) and [a lot of hard work](https://github.com/crsh/papaja/pull/443) by @mariusbarth this later point has come. Adding custom effect sizes including confidence interval was the proof-of-principle case. It is now possible to calculate various effect size measures and report them with their corresponding confidence intervals (calculated from the corresponding test statistics). I will therefore close this PR.

Again, I appreciate your contribution and the interesting discussion about different approaches to obtain confidence intervals for this case.

PR #235, opened 2018-08-01
Proper NA handling for within-subjects CIs and plot functions

mariusbarth:
NAs are now handled explicitly in plot functions and `wsci`(which calculates within-subjects confidence intervals).

- When calculating within-Ss CIs, cases with NAs are removed list-wise (i.e., all observations from this participant are removed), and a warning is issued, it lists the values of the subject identifier of excluded participants.
- Plot functions now have an additional parameter `use`, which allows to include either all non-missing observations (`use = "all.obs"`) or only complete observations (`use = "complete.obs"`) - which, again, implies list-wise removal of participants who provided incomplete data.

### DIFF
--- a/R/calculations.R
+++ b/R/calculations.R
@@ -117,7 +117,24 @@ wsci <- function(data, id, factors, dv, level = .95, method = "Morey") {
     stop("More than one observation per cell. Ensure you aggregated multiple observations per participant/within-subjects condition combination.")
   }
 
-  # split by between factors
+  # ----------------------------------------------------------------------------
+  # Handling of missing values
+
+  if(anyNA(data[[dv]])) {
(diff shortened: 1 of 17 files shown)


R/calculations.R
```diff
@@ -117,7 +117,24 @@ wsci <- function(data, id, factors, dv, level = .95, method = "Morey") {
     stop("More than one observation per cell. Ensure you aggregated multiple observations per participant/within-subjects condition combination.")
   }
 
-  # split by between factors
+  # ----------------------------------------------------------------------------
+  # Handling of missing values
+
+  if(anyNA(data[[dv]])) {
+    excluded_id <- sort(unique(data[[id]][is.na(data[[dv]])]))
```
crsh: See above. This wouldn't catch implicit NAs.

crsh:
Awesome, thanks.

PR #272, opened 2019-03-10
Add how-to-cite info

alexholcombe:


crsh:
Thanks, I probably should have added this information a while ago.

Issue #154, opened 2017-11-02
apa_print for lmer output


mcfrank:
Thanks for all the work on papaja! It would be wonderful if you could `apa_print` merMods, since these models are the standard for much work in psycholinguistics these days.

crsh:
Hi Mike,

thanks for reaching out. Support for hierarchical models is on the top of my wishlist. I'm currently working on methods for `BFBayesFactor` and related objects as well as `lsmobj` and `ref.grid` objects from the `lsmeans` package. Once, those are done I'll start on support for of `lme4` and mixed model objects from `afex`.

If I may, I'll respond to this thread again once I start working on this to get your opinion on the information `apa_print()` should return.

mcfrank:
Thanks, Frederik. I'd be happy to brainstorm what the output should look
like!

jupitus:
Very pleased to see the `BFBayesFactor` table output--thanks! Any update on `lme4` model objects?

crsh:
There are a couple more loose ends to tie up. I'd like to get the first version of `papaja` to CRAN before starting this new feature. I hope, I'll get going on this no later than at the [SIPS 2019 conference](https://www.improvingpsych.org/SIPS2019/). I'm half-planning an ad-hoc hackathon to work on this.

The first and most important question is what information needs to be available for reporting and how it should be formatted (e.g., tables). What a current best practices? Any thoughts, suggestions, and examples are welcome.

khusmann:
I'm very interested in this too! Here's a possible template that I use when reporting results: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2975938/pdf/nihms245694.pdf (see table 2)

crsh:
Great, thank you.

To add to this, here are the papers I know that come closest to "best practice" guidelines:

- [Bolker et al. (2009)](http://max2.ese.u-psud.fr/epc/conservation/PDFs/HIPE/Bolker2009.pdf)
- [Thiele & Markussen (2012)](https://www.uni-muenster.de/imperia/md/content/landschaftsoekologie/oekologischeplanung/team/publikationenjan/potential_of_glmm.pdf)

I also found this preprint, that may be worth looking into: [Meteyard & Davies (2019)](https://psyarxiv.com/h3duq/) with example tables in the [appendix (A5)](https://osf.io/6e8hu/).

khusmann:
[...]

PR #297, opened 2019-07-08
[WIP] apa_print()-methods for lmerMod

kneijenhuijs:
First pass, using broom.mixed on a lmerMod object.

https://github.com/crsh/papaja/issues/154

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,7 +22,9 @@ Imports:
   methods,
   rmarkdown (>= 1.10),
   rmdfiltr,
-  yaml
+  yaml,
+  lme4,
+  broom.mixed
 Suggests:
     tinytex,
     afex,
--- a/R/apa_print_lmerMod.R
+++ b/R/apa_print_lmerMod.R
@@ -1,0 +1,27 @@
+apa.print.lmerMod <- function(
+  lmerMod_fit
+  , conf.level = 0.95
+  , ...
+){
+  
+  tidy_lmerMod_fit <- broom.mixed::tidy(lmerMod_fit, conf.int=TRUE, conf.level=conf.level)
+  
+  tidy_fixed_eff <- subset(as.data.frame(tidy_lmerMod_fit), effect=="fixed")
+  tidy_random_eff <- subset(as.data.frame(tidy_lmerMod_fit), effect=="ran_pars")
+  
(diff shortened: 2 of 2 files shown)


crsh:
Thanks for getting this started, Koen! I'll add a couple of comments, know that this is work in progress.

DESCRIPTION
```diff
@@ -22,7 +22,9 @@ Imports:
   methods,
   rmarkdown (>= 1.10),
   rmdfiltr,
-  yaml
+  yaml,
+  lme4,
```
crsh: I try to keep the mandatory dependencies of `papaja` to a minimum. To do so I add packages for which we provide `apa_print()`-methods to suggests. In this case, it would be necessary to additionally check if `broom.mixed` is installed, because we can't assume that `lme4` users necessarily also use that package. See [here](https://github.com/crsh/papaja/blob/92d5ad4b3740c3b5f8f742f41c7d2584aaf0a3a6/R/calculations.R#L9) for a solution to a similar issue with the `boot` package.

R/apa_print_lmerMod.R
```diff
@@ -0,0 +1,27 @@
...
+  , conf.level = 0.95
+  , ...
+){
+  
+  tidy_lmerMod_fit <- broom.mixed::tidy(lmerMod_fit, conf.int=TRUE, conf.level=conf.level)
+  
+  tidy_fixed_eff <- subset(as.data.frame(tidy_lmerMod_fit), effect=="fixed")
+  tidy_random_eff <- subset(as.data.frame(tidy_lmerMod_fit), effect=="ran_pars")
+  
+  tidy_fixed_eff <- tidy_fixed_eff %>% rename("Effect" = "term"
```
crsh: This is a great start. Maybe a nice next step would be to add variable labels using [`variable_labels()`](https://github.com/crsh/papaja/blob/devel/vignettes/add_apa_print_methods.Rmd#L100).

crsh:
Thanks again for taking a stab at this, @kneijenhuijs. Given that this PR has been stale for some time and @mariusbarth has made a lot of progress on an [alternative implementation](https://github.com/crsh/papaja/pull/395) I'll close this PR. Hope you understand. I very much welcome future contributions from you.

Issue #52, opened 2015-08-26
Add an option for 'table*' to apa_table


saladspices:
To print a two-column table in 'jou' mode, it would be nice to have an option to use `\begin{table*}` instead of `\begin{table}`.  For figures, one can simply use `fig.env='figure*'` in the chunk header, but it seems that there's no such option for tables. Maybe something like:
`apa_table(x, env='table*')`

If there's another option that I'm missing - please let me know!

If you copy the following chunk and the latex code into the default template (changed to 'jou'), you can see the difference:

``` r apa_tab, fig.env='table*'
my_table <- apply(cars, 2, function(x) # Create data
   round(c(Mean = mean(x), SD = sd(x), Min = min(x), Max = max(x)), 2)
)

apa_table(
   my_table
   , align = c("l", "r", "r")
   , caption = "A summary table of the cars dataset."
   , note = "This table was created using apa\\_table()"
   , added_colnames = "Descriptives"
)
```

``` r
\begin{table*}
\begin{center}
\begin{threeparttable}
\caption{A summary table of the cars dataset.}
\begin{tabular}{lrr}
\toprule
Descriptives & \multicolumn{1}{c}{speed} & \multicolumn{1}{c}{dist}\\
\midrule
Mean & 15.4 & 42.98\\
SD & 5.29 & 25.77\\
Min & 4 & 2\\
Max & 25 & 120\\
\bottomrule
\end{tabular}
\begin{tablenotes}
\textit{Note.} This table was created using apa\_table()
\end{tablenotes}
\end{threeparttable}
\end{center}
\end{table*}
```

crsh:
If I understand correctly you want to use `table*` to place the table in the center of an otherwise empty page. Is that correct? A quick fix, which I know is not quite what you want, would be to set `placement = "p"` when you call `apa_table.latex()`.

In principle, adding an option like you describe seems straight forward. I'm a little hesitant because the function uses different nested environments depending on options such as `landscape` or `longtable`. I need to explore the possible side effects of the combinations of these options.

crsh:
Okay, this is a little tricky and will require some more experimenting on my part. `longtable` doesn't work in `jou` mode at all, so I'll need to give this some more thought and it will require some rewriting of `apa_table()`.

Another hack I just thought of that should help is the following:
~~~{r}
pretty_table <- capture.output(apa_table(
   my_table
   , align = c("l", "r", "r")
   , caption = "A summary table of the cars dataset."
[...]

PR #298, opened 2019-07-08
Update apa_table #52

jstbcs:
Changing default table environment from table to table*. This changes the default behavior when class: "jou" so that tables spread two columns.

### DIFF
--- a/R/apa_table.R
+++ b/R/apa_table.R
@@ -490,7 +490,7 @@ apa_table.latex <- function(
 
   # if(longtable && placement != "h") cat("\\afterpage{\\clearpage") # Defer table to next clear page
   # table_output <- c(table_output, "\n\n")
-  if(!landscape && !longtable) table_output <- c(table_output, paste0("\\begin{table}", place_opt))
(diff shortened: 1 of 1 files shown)


crsh:
Hi Julia, thanks for jumping on this. I think this is a big step towards better support for `jou`-documents! I think your approach is right on target. From what I understand this default-change will have [two consequences in terms of placement](https://tex.stackexchange.com/a/39020/11001): The table will always be deferred and it's not possible to place tables at the bottom of a page. I think both could be confusing in `man`-mode. Would you be willing to ammend this PR to change this default behavior only if the document is knitted with in `jou` mode? A straight forward way to check this would be `any(grepl("jou", c(rmarkdown::metadata$classoption, rmarkdown::metadata$class)))`.

crsh:
Thanks again for the PR! I'm closing this in favor of https://github.com/crsh/papaja/commit/c0962eb3115918655575d85c77e602f452273f84, which is a slightly more elaborate implementation that takes the `classoption` field into but allows it to be overwritten to allow small tables to be single-column.

PR #299, opened 2019-07-08
Metafor

ekothe:
This is the first attempt at adding a apa_print function for metafor (rma) objects. This will handle both univariate meta-analysis and meta-regression models. 

The formatting of the output and documentations needs improvement but this should give you an idea of where we're up to after the hackathon at SIPS2019.


R/apa_print_metafor.R
```diff
@@ -0,0 +1,57 @@
...
+#'      \item{\code{estimate}}{A named list of character strings giving the descriptive estimates and confidence intervals
+#'          for each term.}
+#'      \item{\code{full_result}}{A named list of character strings comprised of \code{estimate} and \code{statistic} for each term.}
+#'      \item{\code{table$estimate}}{A data.frame containing the study label, estimate, standard error, test statistic for each observed study and the overall meta-analytic estimate with standard error, test statistic, and p value. This table can be passed to \code{\link{apa_table}}.}
+#'    }
+#'    #'      \item{\code{table$heterogeneity}}{A data.frame containing the number of observations, measure and method employed in the analysis as well as the I^2, Tau^2 Tau^2 standard error and results of Cochran test of heterogeneity. This table can be passed to \code{\link{apa_table}}.}
+#'    }
+#' @family apa_print
+#'
+apa_print.metafor <- function(
```
crsh: Unless I'm missing something, for S3-method dispatch to work this functions needs to be named after the S3 class it is ment to operate on. In this case, I think, it should be either the specific `rma.uni` or more general `rma` depending on how general this method should be. Right?

R/apa_print_metafor.R
```diff
@@ -0,0 +1,57 @@
...
+){
+
+  if(!is(x, "rma")) stop("Object is not an rma object")
+  if(!is.null(est_name)) papaja:::validate(est_name, check_class = "character", check_length = 1)
+  if(is.null(est_name)) est_name <- "b"
+
+
+  apa_res <- papaja:::apa_print_container()
+
+  apa_res$estimate <- paste0("$", est_name, "$ = ", round(x$b, x$digits[["est"]]), " $95\\% CI$ = $[", round(x$ci.lb, digits = x$digits[["ci"]]), "$, $", round(x$ci.ub, digits = x$digits[["ci"]]), "]$")
```
crsh: In terms of a general workflow [I try to assemble the `table` element first](https://github.com/crsh/papaja/blob/devel/vignettes/add_apa_print_methods.Rmd#L194) and build the rest from it. Do you think this could be a sensible approach here too? It may simplify maintenance and may expose commonalities among functions that could be used further reuse code across methods.

R/apa_print_metafor.R
```diff
@@ -0,0 +1,57 @@
...
+  apa_res$estimate <- setNames(apa_res$estimate, row.names(x$beta))
```
crsh: Nice, I never use `setNames()` but really should.

R/apa_print_metafor.R
```diff
@@ -0,0 +1,57 @@
...
+  apa_res$statistic <- paste0("$z$ = ", round(x$zval, digits = x$digits[["test"]]), ", $p$ = ", printp(x$pval))
+  apa_res$statistic <- setNames(apa_res$statistic, row.names(x$beta))
+
+  apa_res$full_result <- paste(apa_res$estimate, apa_res$statistic)
+  apa_res$full_result <- setNames(apa_res$full_result, row.names(x$beta))
+
+  apa_res$table$estimate <- broom::tidy(x)
+
+  apa_res$table$heterogeneity <- broom::glance(x)
+  names(apa_res$table$heterogeneity) <- c("k", "measure", "method", "I$^2$", "H$^2$", "$\\tau^2$", "$\\tau^2$ se", "Cochran Q$_e$", "Cochran Q$_e$ $p$", "Cochran Q$_m$", "Cochran Q$_m$ $p$")
```
crsh: Maybe a nice next step would be to add variable labels using [`variable_labels()`](https://github.com/crsh/papaja/blob/devel/vignettes/add_apa_print_methods.Rmd#L100).

R/apa_print_metafor.R
ekothe: To be honest it's just copied from how the variable names are set in metafor itself.

crsh:
I hate to see this PR go to waste, but since this PR was opened papaja has seen so many changes, that support for **metafor**, which I would love to see, would probably be best implemented from scratch. Hence, I'm closing this PR hoping that someone will give this another shot.

PR #309, opened 2019-08-27
Adds simple codebook function

crsh:
Adds a function that creates a very simple codebook in CSV format containing variable names, labels, class, and some basic descriptive summaries.


crsh:
Yeah, that's much better.

Issue #233, opened 2018-07-30
Support MANOVA reporting in apa_print.aov_ez()


crsh:
`afex` supports calculating [multivariate tests](https://afex.singmann.science/forums/topic/manova) for each effect. It would be great if it was possible to report these with `apa_print()`.

mariusbarth:
I checked what we would have to do to make this possible: Basically, `aov_ez` objects contain the multivariate tests provided by class `Anova.mlm` from the `car` package. The multivariate statistics are computed within the `print` method and are never returned in a proper output object by `car` (so this is the same problem as with MANOVA objects from `car`). A possible solution would be to use the functions used to calculate multivariate statistics (which seem to be copied from the `stats` package). Unfortunately, these are not exported, so I think we would have to copy them. Still, because Type 3 sums of squares are ubiquitous in psychology and these are virtually always calculated using `car`, I would definitely consider it, especially because the code is very simple and only relies on `base` functions.

crsh:
Thanks for looking into this, Marius! I remember that the essential calculations in the print method had been a road block for these things. Copying the code would be the quick fix option and I'm generally fine with this as an interim solution. But it may be worth proposing a contribution to the `car` where, at the least, the function that performs these calculations is exported.

PR #310, opened 2019-08-27
Add support for `manova` and `summary.manova` for `apa_print()`

mariusbarth:
Adds support for `manova` and `summary.manova` for `apa_print()`. Only MANOVA objects from the `stats` package, not from the `car` package, are supported. `car` does not return the necessary test statistics (these are calculated within `print.Anova.mlm` and are thus not accessible).

- In addition, now an error is returned if an object of class `Anova.mlm` that does not contain univariate tests (which is the case for MANOVA output) is handed to `apa_print()`
- Moreover, I added a new utility function, `add_equals()`: Here we should think about adding `"\\geq"` and `"\\leq"` to the patterns to be matched.
- The `apa_print()` methods for MANOVA only work with the newest development version of `broom` installed, yet.


R/apa_print_manova.R
```diff
@@ -0,0 +1,115 @@
...
+  estimate_names_with_dollars <- paste0("$", estimate_names, "$")
+  names(estimate_names_with_dollars) <- names(estimate_names)
+
+  multivariate_stat <- intersect(names(estimate_names), colnames(x))
+
+  if(length(multivariate_stat) == 0L) {
+    stop("No multivariate test statistic found in results object.")
+  } # move this to input validation
+
+  col_names <- c(
```
crsh: Hmm, I'm not sure about this bit. I can see you followed the behavior of `print_anova()` here. However, I wonder whether it would make more sense to keep the naming of the columns consistent with the `broom` output. I think this is what we do for regression, too. It would probably make sense to tackle this in a separate PR, but I'd be interested in your opinion on this.

R/apa_print_manova.R
```diff
@@ -0,0 +1,115 @@
...
+    , "df2" = "$\\mathit{df}_2$"
+    , "p" = "$p$"
+  )
+
+  x <- x[x$Effect != "Residuals", colnames(x) != "df", drop = FALSE]
+
+  x[["Effect"]] <- prettify_terms(x[["Effect"]])
+  x[[multivariate_stat]] <- printnum(x[[multivariate_stat]])
+  x[["F"]] <- printnum(x[["F"]])
+  x[["df1"]] <-as.character(round(x[["df1"]], digits = 2))
```
crsh: Out of curiosity: Why don't you use `printnum()` here?

R/apa_print_manova.R
```diff
@@ -0,0 +1,115 @@
...
+  res$statistic <- as.list(apply(X = x, MARGIN = 1, function(y) {
+    stat <- paste0(
+      "$", estimate_names[[multivariate_stat]], " = ", y[[multivariate_stat]], "$, "
+      , "$F(", y[["df1"]], ", ", y[["df2"]], ") = ", y[["F"]], "$, "
+      , "$p ", add_equals(y[["p"]]), "$"
+    )
+    stat
+  }))
+  if(in_paren) res$statistic <- in_paren(res$statistic)
+  names(res$statistic) <- gsub(x$Effect, pattern = " $\\times$ ", replacement = "_", fixed = TRUE)
```
crsh: Any particular reason you are not using `sanitize_terms()` on the original effect names here?

R/arrange_anova.R
```diff
@@ -135,6 +135,14 @@ arrange_anova.summary.aovlist <- function(x) {
 arrange_anova.summary.Anova.mlm <- function(x, correction = "GG") {
   validate(correction, check_class = "character", check_length = 1)
 
+  # univariate.tests is NULL if the object comes from a MANOVA
+  if(is.null(x$univariate.tests)) {
+    stop(
+      "Anova.mlm objects from car::Manova are not supported, yet. "
```
crsh: Could you add something about requesting this on GitHub? Something along the lines of the output of `no_method()` maybe.

R/utils.R
```diff
@@ -567,3 +567,26 @@ complete_observations <- function(data, id, within, dv) {
...
+#' This is an internal function that prepends every element of a character
+#' vector with an 'equals' sign if the respective element does not contain one of
+#' \code{c("=", "<", ">")}.
+#'
+#' @param x A character vector.
+#'
+#' @keywords internal
+
+
+add_equals <-function(x) {
```
crsh: Nice. Would you be willing to search the code base for other instances where this could be used?

crsh:
Maybe one more question: Do you think it would make sense to put the MANOVA-methods on a separate documentation page?

crsh:
Sorry, one more thing: `afex` supports multivariate tests as well (see https://github.com/crsh/papaja/issues/233). Have you looked at whether these can be supported as well? Would you be willing to do this within this PR or would you prefer to do this in a separate one?

R/utils.R
mariusbarth: Sure.

R/apa_print_manova.R
mariusbarth: I'm not aware of any correction methods for degrees of freedom in MANOVA, but this way the default is the same as for ANOVA: If integer df are provided, no decimals are reported; if there are corrected ones, they will be reported rounded up to two digits (edge cases where only one digit is significant, but two should be printed are not captured, yet).

R/apa_print_manova.R
mariusbarth: not aware of sanitize_terms... :laughing:

R/apa_print_manova.R
mariusbarth: You mean term, statistic, and num.df, den.df instead of Effect, F, df1, and df2?

1. I think it would be very confusing to coin a column "term" and then label it "Effect".
2. I think the output from broom here is quiet inconsistent: coining one column "pillai" and the other "statistic" is less consistent than using "pillai" and "F" (it would also be  to use "multivariate.statistic" and "univariate.statistic")

mariusbarth:
> Maybe one more question: Do you think it would make sense to put the MANOVA-methods on a separate documentation page?

I think this would result in a mess if we later supported `Anova.mlm` objects for MANOVA -- two methods of the same function documented in different Rd files. Can roxygen handle cases like this?

R/arrange_anova.R
mariusbarth: Sure.

mariusbarth:
>Sorry, one more thing: afex supports multivariate tests as well (see #233). Have you looked at whether these can be supported as well? Would you be willing to do this within this PR or would you prefer to do this in a separate one?

Separate PR: I thought the multivariate tests are only provided by means of the car package, which would result in the same problems as with car itself (calculations hidden in print methods), but I could check if the other output elements also allow for calculating multivariate tests

R/apa_print_manova.R
crsh: Given that you implemented the MANOVA methods in `broom`, we know who's to blame for the inconsistency here. :stuck_out_tongue_winking_eye:

I understand you concern that indexing with `$term` although the column is printed with the heading `Effect` may be confusing. I suggest we think this through a little more. I think it's definitely desirable to be as consistent as possible in our naming of `apa_result_table`-objects.

R/apa_print_manova.R
crsh: Ah, I see. The downside of this approach is that any df that come out as having exactly 1 digit will be printed with only one digit. How about something like this:

~~~r
df <- c(1, 2, 1.1, 1.35987, 5.210938)
printnum(df, digits = as.numeric(df %% 1 != 0) * 2)
~~~

crsh:
> I think this would result in a mess if we later supported `Anova.mlm` objects for MANOVA -- two methods of the same function documented in different Rd files. Can roxygen handle cases like this?

I see. I'm not sure. A quick Google search yielded no results. At the very least this can probably be done using the `@eval` tag.

> Separate PR: I thought the multivariate tests are only provided by means of the car package, which would result in the same problems as with car itself (calculations hidden in print methods), but I could check if the other output elements also allow for calculating multivariate tests

As we just saw, you are right. Again this would require reimplementing some of the computations currently done in the print method. That's a dealbreaker.

crsh:
Thanks! :)

PR #325, opened 2019-10-24
Update skeleton.Rmd

setgree:
Added `papaja::r_refs('r-references.bib')` to the first code chunk. Idea here is that when I try to compile this from scratch, the first time, it warns me that r-references.bib is missing, and if you're using a container to execute this as a one-off (e.g. Code Ocean), it'll be a problem each time. So the point is to create the file before it is called in line 84.

### DIFF
--- a/inst/rmarkdown/templates/apa6/skeleton/skeleton.Rmd
+++ b/inst/rmarkdown/templates/apa6/skeleton/skeleton.Rmd
@@ -59,7 +59,7 @@ output            : papaja::apa6_pdf
 
 ```{r setup, include = FALSE}
 library("papaja")
-papaja::r_refs("r-references.bib")
+r_refs("r-references.bib")
 
 ```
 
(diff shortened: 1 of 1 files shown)


crsh:
Hi Seth, sorry this has been a really long time for such a tiny change. What you are suggesting makes sense.  I only have some cosmetic suggestions and I'll be happy to merge this PR. Thanks for your hard work on making `papaja` available so conveniently to CodeOcean users!

inst/rmarkdown/templates/apa6/skeleton/skeleton.Rmd
```diff
@@ -59,6 +59,8 @@ output            : papaja::apa6_pdf
 
 ```{r setup, include = FALSE}
 library("papaja")
+papaja::r_refs("r-references.bib")
```
crsh, line 62: Given that `papaja` is loaded above, we don't really need the `papaja::` prefix here. Also, could you add an empty line between the library call and the call to `r_refs()`?

crsh: Sorry, one more thing: Could your remove the chunk in lines 95-97, now that the reference file is created up here?

setgree:
Hi @crsh , thanks so much! And I apologize for not getting to implementing the changes you suggested -- everything looks great to me, thanks again for attending to it.

PR #367, opened 2020-05-13
Assorted bug fixes

mariusbarth:
Hi @crsh,

I worked on support for hierarchical (generalized) linear models, which I think already works nicely. I think there are also some loose ends that need some discussion and design decisions. However, I'm filing this pull request because along the way I also fixed some bugs that now occurred on Travis and also on my laptop where I run tests with R 4.0.0. I thought it may be worthwhile to merge all these fixes to the development branch before continuing (and also to address #361). I know you're super busy, but thought you may be convinced to merge quickly into the development branch. 

Here's a summary of what I changed:
- added `apa_print()` methods for objects of class `merMod` and `mixed`.
- Because `mixed` objects basically contain a table of class `anova`, I worked on making `print_anova` and `arrange_anova` more general: Both are now capable of dealing with columns that may be present in some objects but not others, and also to deal with the column names typically found in analysis of deviance tables (chi squared, only one df).
- I fixed the namespace problem mentioned in #361. Methods for `apa_print()` are now properly exported, again.
- In the model-comparison table, each column of the data.frame used to be a named vector in R 3.6., but an unnamed one in R 4.0, which resulted in a failing test. I asserted that unnamed vectors are intended behaviour, because the names are already stored in the `row.names` attribute of the `data.frame`: Therefore, I ensured that the extraneous names are removed in both 3.6 and 4.0.
- I also removed `report = "warn"` from `rmdfiltr::add_wordcount_filter(args, report = "warn")` without thinking about it.
- "HF eps > 1 treated as 1." is now a message, not a warning (we already talked about this one a while ago)
- Fixed the docs of `printp()`
- `variable_label()` is now more verbose when trying to assign a label to a non-existing column: It not only complains, but also tells the user which column cannot be found.
- I removed support for objects of class `summary.manova`, because the respective tidy-method in broom is not available from CRAN, yet.
- I now more strictly test the apa_print() generic: The test checks whether the string "github.com/crsh/papaja/issues " is present in the error message: With this test, we would have caught the namespace problem in #361.
- I (temporarily) added my fork of broom to `Remotes` in the `DESCRIPTION` file, because installing from `tidymodels/broom` is currently not possible. Filed a pull request there, but wanted Travis to work quickly, again.


mariusbarth:
P.S.: What do you think about a version bump?

R/apa_print.R
```diff
@@ -32,4 +32,7 @@ apa_print <- function(x, ...) {
 # #' @importClassesFrom BayesFactor BFBayesFactor
 setGeneric("apa_print")
 
+#' @method apa_print default
```
crsh, line 35: Hmm, I seem to remember reading in the `roxygen` documentation that the methods generally do not need to be specified manually. But I assume this is somehow necessary due to recent issues with method dispatch?

R/apa_print_anova.R
```diff
@@ -133,11 +141,31 @@ apa_print.anova <- function(
...
+  # Some model objects do not contain the necessary information to calculate
+  # MSEs and effect-size measures. In such cases, print_anova() needs to be called
+  # with arguments es = NULL and mse = FALSE.
+
   if("apa_variance_table" %in% class(variance_table)) { # car::LeveneTest
     if(length(object_heading) == 1 && grepl("Levene", object_heading)) {
       if(!is.null(ellipsis$es)) stop("Effect sizes are not available for car::LeveneTest-objects.")
       return(print_anova(variance_table, es = NULL, mse = FALSE, ...))
     }
+    # lmerTest::anova.merModLmerTest ----
```
crsh, line 153: Could you comment out these lines for the time being so that they show up in the next PR for me to review?

R/arrange_anova.R
```diff
@@ -60,21 +60,73 @@ arrange_anova.anova <- function(x) {
...
+      object$`Chi Df` <- NULL
+      object$`Pr(>Chisq)` <- NULL
+    }
+    if("Chi Df" %in% col_names && "Df" %in% col_names) {
+      object$Df <- NULL
+    }
 
-  } else if(is.null(object[["Sum Sq"]])) { # car::levenTest
-    x <- x[, -which(colnames(x) %in% c("sumsq", "sumsq_err"))]
+    renamers <- c(
```
crsh, line 84: Do we maybe want to put this lookup table in a place of its own? I suspect this may be applicable more generally, no? Seems like a similar use case as [`convert_stat_name()`](https://github.com/crsh/papaja/blob/b94783a7c62643abfeeee57c6008622709378306/R/utils.R#L153).

R/print_anova.R
```diff
@@ -79,63 +79,89 @@ print_anova <- function(
...
 
   if(!is.null(correction_type) && correction_type != "none") {
     variable_label(anova_table) <- c(
-      "Effect" = "Effect"
-      , "F" = "$F$"
-      , "df1" = paste0("$\\mathit{df}_1^{", correction_type, "}$")
-      , "df2" = paste0("$\\mathit{df}_2^{", correction_type, "}$")
-      , "p" = "$p$"
-    )
+      "Effect"  = "Effect"
```
crsh, line 112: Similar to the above. Do you think we can turn these into generally applicable lookup tables that we place in their own utility functions?

crsh:
One more thing. Could you maybe already summarize your changes in the NEWS file?

R/apa_print.R
mariusbarth: Yes, this is what I also remembered. I searched for the source, it was Hadley Wickham's R packages book... however, end of last year he added a paragraph that disambiguation might be necessary and then `@method` should be used: http://r-pkgs.had.co.nz/man.html#man-classes
There also seems the something in the making with `@exportS3method`, but I could not find which version of `roxygen2` already supports it. I thought going with the stable version would be wise, especially because there seems to be changing so many thing at once, right now.

mariusbarth: Short answer: Yes, it's necessary. :-D

R/apa_print_anova.R
mariusbarth: Sure.

R/arrange_anova.R
mariusbarth: Love the idea! We could make the process more robust (e.g., by using `make.names()`) and transparent (one table for all, so it's like a dictionary). The only pro for separate lists would be that you can omit some column names that are not legal in a specific `apa_print()` method

However, I think a renaming function could take an argument with a whitelist of legal column names (that we anticipated while programming), maybe throwing a warning, not an error, when there are undetected columns.

R/print_anova.R
mariusbarth: I think for this lookup table it won't work, because we are going from the generic term to a method-specific term. But I think the process of renaming columns (last comment) and adding variable labels (this comment) might be done within one step, with the labels created from the original column name (or at least a best guess that may be altered within a specific `apa_print()` method). This way, we should be able to make this step rock solid.

R/apa_print.R
crsh: If I understand correctly, `@exportS3method` was introduced in [`roxygen2` 7.0.0](https://github.com/r-lib/roxygen2/blob/4b501fe3e6ea8ca4e630f108d1ba8c84c55ac601/NEWS.md#roxygen2-700). So we should be able to use it, right?

R/arrange_anova.R
crsh: Yes, that was my thinking too. So we can have lookup methods for renaming columns, for labelling columns and so on. If there are any ambiguities we can still resolve those in place, I think.

R/print_anova.R
crsh: I see. If you can get it to work that way that would be great. Otherwise we can just keep it like this for now.

R/apa_print.R
mariusbarth: Didn't work on my machine, so I avoided it. :/

R/print_anova.R
mariusbarth: I think we should keep it like that, right now, and I will try to file a new PR with a possible implementation, soon.

R/apa_print.R
mariusbarth: I tried out `@exportS3Method` which now works (it was the capitalization that played a trick on me). Still, the effect is exactly the same as using `@method` and `@export` together. The only advantage of `@exportS3Method` seems to be that you are now able to write methods for generics from other packages more easily by using `@exportS3Method pkh::generic class`.

R/apa_print_anova.R
mariusbarth: Did that.

R/apa_print.R
crsh: Nice. So lets use `@exportS3Method` as it is the recommended approach and has a technical advantage.

crsh:
@mariusbarth Just checking in: What's the state of things here. Would you like another round of review or are you still working on this?

mariusbarth:
Öhm, I'm waiting for you to merge. :rofl: I think I addressed every issue raised in your review

R/apa_print.R
mariusbarth: Would you mind merging this PR first and then I commit this change to the development branch in the main repository? My fork is already filled with many other things. :joy:

crsh:
Okay, great. Next time just drop me a quick comment, so I know when you are done making changes. :)

R/apa_print.R
crsh: Sure.

R/arrange_anova.R
crsh: Not sure if I'm missing something but is this addressed somewhere?

mariusbarth: I implemented a new function `sanitize_table()` with lookup methods for both column names and labels on another branch which already works nicely with `htest` and `manova` objects; however, I thought it is wise to hold this back for another PR, because it might still be necessary to change the workflow a bit when we start using it throughout the package. Moreover, before deploying such strongly altered code, I wanted to add some more rigid testing.

crsh: Alrighty, I hope you're keeping a list of all the other PRs you have in the pipeline. 😆

mariusbarth:
So I resolved the remaining merge conflicts and successfully ran all tests and checks. If you want, feel free to merge (maybe take a final look at `apa6_formats.R` and check if I did anything wrong).

crsh:
Great, thanks!

PR #369, opened 2020-05-18
Include cslreferences class in revision_letter template

jvcasillas:
This change includes the cslreferences change implemented in pandoc 2.8.


crsh:
Thanks!

PR #385, opened 2020-07-10
Closing unused connection message

mutlusun:
Hello,
adding a appendix file to my papaja project resulted always in the following two messages:
```
2: closing unused connection 5 (appendix.tex) 
3: closing unused connection 4 (PATH/appendix.md)
```

This PR correctly closes these files. I could not use the `closeAllConnections` function because of an error. I assume that `closeAllConnections` closes a connection that is still used/needed (maybe from the calling function/environment). I checked the rest of the code whether `on.exit` is called twice without `add=TRUE`. That's not the case (yet).

Best


crsh:
Looks good. Would you like to add yourself as a contributor in `DESCRIPTION` before I merge this PR?

mutlusun:
Thanks for your feedback! Done!

crsh:
Excellent, thanks for your contribution and thank you @mariusbarth for taking a look.

PR #390, opened 2020-07-17
Printnum optimizations

mariusbarth:
I'm proud to announce some modifications to `printnum()` that will speed up code execution throughout the whole package. Basically, I streamlined the code by retiring `printnumber()` and doing all formatting directly within `printnum.numeric()` with vector-valued functions (whenever possible).

The only part that cannot be vectorized easily is the call to `formatC()`, because `formatC()` does not accept vector-valued arguments. I built a workaround that first checks whether or not the arguments to `formatC()` are vector-valued; if this is the case, `formatC()` is called repeatedly on each element of `x`. If all arguments (except `x`) to `formatC` are of length one, `formatC()` is applied to the vector of `x`.

These changes considerably speed up `printnum()`, with a minimum speedup of ~1.2 (when `length(x)==1`), but increasing sppedups with vector length (see appended figure, left the case when all args to formatC have length 1, right if this is not the case).
grafik
grafik


R/printnum.R
```diff
@@ -16,9 +16,9 @@
 #' @param zero_string Character. Word to print if \code{x} is a zero integer.
 #' @inheritDotParams base::formatC -x
 #'
-#' @details If \code{x} is a vector, \code{digits}, \code{gt1}, and \code{zero} can be vectors
-#'    according to which each element of the vector is formated. Parameters are recycled if length of \code{x}
-#'    exceeds length of the parameter vectors. If \code{x} is a matrix, the vectors specify the formating
+#' @details If \code{x} is a vector, all arguments can be vectors
+#'    according to which each element of the vector is formatted. Parameters are recycled if length of \code{x}
+#'    exceeds length of the parameter vectors. If \code{x} is a matrix or data.frame, the vectors specify the formating
```
crsh, line 21: Could you add code tags to "matrix" and "data.frame" and add a second "t" to "formating" while you are at it?

R/printnum.R
```diff
@@ -302,6 +345,111 @@ printnum.papaja_labelled <-function(x, ...){
...
+#' @examples
+#' printp(0.05)
+#' printp(0.0005)
+#' printp(0.99999999)
+#' printp(c(.001, 0), add_equals = TRUE)
+#' @export
+
+printp <- function(x, digits = 3L, na_string = "", add_equals = FALSE) {
+  validate(x, check_class = "numeric", check_range = c(0, 1), check_NA = FALSE)
+  validate(na_string, check_class = "character")
```
crsh, line 364: I think the `na_string`- and `add_equals`-validation here is redundant with the checking done in `printnum()`, right?

crsh:
This looks great. Two quick questions:

1. I'm not sure I understand what's happening with the `margin` argument. From what I can tell, it's still in the documentation but removed from `printnum()` (including validation).
2. What's `printnum_old()` for? As far as I can tell, it's not being called anywhere, right?

mariusbarth:
>     1. I'm not sure I understand what's happening with the `margin` argument. From what I can tell, it's still in the documentation but removed from `printnum()` (including validation).

I removed it from the function arguments because it didn't have any effect in the vector-case, anyways (this argument was a relic from the times when `printnum()` was not an S3 generic, yet). However, to get some extra safety, I explicitly remove it from the ellipsis, just for those cases when someone passed the `margin` arguemnt to `printnum()` and it now (with a newer version of `papaja`) dispatches to the `numeric` method. Just ensuring it is not passed to `formatC`, which would result in an error.
The `margin` argument is still in the documentation, because it's still available for the `matrix` and `data.frame()` methods.

>     2. What's `printnum_old()` for? As far as I can tell, it's not being called anywhere, right?

`printnum_old()` is only our old `printnum.numeric()` that I left in the code for comparing performance to the new function. I think we can safely remove it (the same applies to `printnumber()`).

R/printnum.R
mariusbarth: Yes. It would be my pleasure and privilege to remove these extraneous validation steps.  :laughing:

R/printnum.R
mariusbarth: Done.

R/printnum.R
mariusbarth: Done.

crsh:
Thanks, that makes sense. So, if we don't need `printnum_old()` and `printnumber()` feel free to delete them. Other than that, the PR is ready to be merged. :+1:

mariusbarth:
Ok, I will remove them and then merge.

PR #395, opened 2020-07-31
Support hierarchical (generalized) linear models from lme4, lmerTest, and afex

mariusbarth:
Hi @crsh,
this PR adds support for hierarchical (generalized) linear models from lme4, lmerTest, and afex to `apa_print()`, using our new processing pipeline. There is one issue that came to my mind:

- The `apa_print.anova()` method became even more complicated than it already used to be. I think this is best solved by moving the old ANOVA parts to the new pipeline.


crsh:
> The apa_print.anova() method became even more complicated than it already used to be. I think this is best solved by moving the old ANOVA parts to the new pipeline.

I think this is desirable anyway. We should try to refactor all existing methods to the new workflow, if possible. But let's leave this for a separate PR.

mariusbarth:
> I think this is desirable anyway. We should try to refactor all existing methods to the new workflow, if possible. But let's leave this for a separate PR.

Sure

mariusbarth:
Ok, I think now I'm done. :smile:

R/glue_apa_results.R
```diff
@@ -71,7 +72,7 @@ glue_apa_results <- function(x = NULL, ...) {
     )
 
     if(!is.null(x) && is.data.frame(x)) {
-        if(!is.null(x$term)) x$term <- prettify_terms(x$term)
+        # if(!is.null(x$term)) x$term <- prettify_terms(x$term) BUGFIX: applied prettify_terms() twice
```
crsh, line 75: I understand the function is applied twice with your workflow, but if we go through `broom`, which I would still prefer, it would still need to be applied here, no?

R/utils.R
```diff
@@ -346,18 +346,17 @@ sanitize_terms <- function(x, standardized = FALSE) {
 }
 
 
-#' Prettify term names
+#' Prettify Term Names
```
crsh, line 349: Do we really want to export this function at this point? Before exporting, I think we should, at the very least, reconsider the name of this function. I think 'prettify'  is not very descriptive.

R/glue_apa_results.R
mariusbarth: I also wondered what might be the best way to create a consistent workflow here. I concluded that the internal function `beautify()` does typesetting of all columns of the `apa_results_table`, and it should therefore also typeset the `$term` column. A similar line of thought applies to `glue_apa_results()`: In the examples section, you manually typeset the columns that are not processed with the standard `printnum_transformer`, and I think this should also happen with the `$term` column, which is why I exported `prettify_terms()`. However, I also believe that the function name really is not perfect, and I would opt for something like `print_terms()` (because it does typesetting like `printnum()`, `print_df()` or `printp()`). As an alternative, maybe it's a good idea to include default typesetting facilities in `glue_apa_results()` (e.g. guessing which one of the functions `printp()`, `printnum()`, `print_df()` or `print_terms()` should be applied to each column). If we provided these facilities, we could refrain from exporting `prettify_terms()`

R/utils.R
mariusbarth: see above

R/glue_apa_results.R
crsh: Would it be possible to revert this change, so we can have this discussion in a separate issue/PR?

mariusbarth: Sure, good point.

R/apa_print_anova.R
```diff
@@ -137,34 +137,106 @@ apa_print.anova <- function(
...
-    #     c(KR = "KR", S = "S", PB = "none", LRT = "none")[attr(x, "method")]
-    #   )
-    #   return(print_anova(variance_table, mse = FALSE, es = NULL, ...))
-    # }
-    # # lmerTest::ranova ----
-    # if(object_heading[1] == "ANOVA-like table for random-effects: Single term deletions") {
-    #   stop("Single-term deletions are not supported, yet.\nVisit https://github.com/crsh/papaja/issues to request support.")
-    # }
+
+    # lmerTest::anova.merModLmerTest -------------------------------------------
```
crsh, line 172: There seems to be some avoidable redundancy here (setting attributes, creating `Effect` column and  removing rownames, call to `canonize()`, `beautify()`, and `glue_apa_results()`).  Could we add one nesting `if` layer to avoid this?

R/glue_apa_results.R
```diff
@@ -31,6 +31,7 @@
 #' iris_lm <- lm(Sepal.Length ~ Petal.Length + Petal.Width, iris)
 #' tidy_iris_lm <- broom::tidy(iris_lm, conf.int = TRUE)
 #' tidy_iris_lm$p.value <- printp(tidy_iris_lm$p.value)
+#' tidy_iris_lm$term <- tidy_iris_lm$term
```
crsh, line 34: I this really what you want here? ;)

R/utils.R
```diff
@@ -205,30 +205,58 @@ canonize <- function(
...
     if(!is.null(conf_level)) paste0(conf_level * 100, "\\% ")
     , "CI"
   )
-  # print(conf_label)
+
+  # Corrections of df from
+  # - Hierarchical Linear Models: Kenward-Roger and Satterthwaite
+  # - Repeated-measures ANOVA   : Greenhouse-Geisser and Huyhn-Feldt
+  correction_type <- attr(x, "correction")
+  if(is.null(correction_type)) {
```
crsh, line 224: Wouldn't it be easier to paste `^{...}` in case of a non-null `correction_type`?

R/utils.R
```diff
@@ -205,30 +205,58 @@ canonize <- function(
...
+    , "denDF"        = label_df2
   )
 
   names_in_lookup_names <- colnames(x) %in% names(lookup_names)
 
+
+  warning_unexpected <- "\nThis implies that your output object was not fully understood by `apa_print()`.
+  Therefore, be careful when using its output. Moreover, please visit https://github.com/crsh/papaja/issues and
+  file an issue together with the code that generated the output object. In doing so, you help us to fully
+  support the type of analysis you just conducted and make papaja a lttle bit better."
```
crsh, line 256: `lttle` -> `little`

R/utils.R
```diff
@@ -280,7 +307,11 @@ beautify <- function(x, standardized = FALSE, use_math = FALSE, ...) {
 
   # rearrange ----
   multivariate <- paste0("multivariate.", c("statistic", "df1", "df2"))
-  ordered_cols <- intersect(c("term", "estimate", "conf.int", multivariate, "statistic", "df", "df1", "df2", "p.value"), colnames(x))
+
+  se <- NULL
+  if(!any(colnames(x) == "conf.int")) se <- "std.error"
+
+  ordered_cols <- intersect(c("term", "estimate", "conf.int", se, multivariate, "statistic", "df", "df1", "df2", "p.value"), colnames(x))
```
crsh, line 314: The order of columns should be standardized across methods, right? Could we turn this into a convenience function?

R/variable_label.R
```diff
@@ -122,7 +122,7 @@ assign_label.data.frame <- function(x, value, ...){
   }
 
   for(i in seq_along(colnames(x))) {
-    if(colnames(x)[i] %in% names(value)) {
+    if(any(colnames(x)[i] == names(value))) {
```
crsh, line 125: Just out of curiosity, what's the advantage of this approach?

R/apa_print_merMod.R
```diff
@@ -1,144 +1,112 @@
...
+#' @rdname apa_print.merMod
+#' @method apa_print merMod
+#' @export
+#' @md
+
+apa_print.merMod <- function(
+  x
+  , args_confint
+  , in_paren = FALSE
+  , est_name
```
crsh, line 29: In most other methods we define `est_name = NULL` as default. Any particular reason to deviate from this here? I like the `NULL`-default because it makes explicit, that no value needs to be passed.

crsh: Same for `args_confint`.

R/apa_print_merMod.R
```diff
@@ -1,144 +1,112 @@
...
+
+  # Rearrange ----
+  x_summary <- summary(x)
+
+  res_table <- as.data.frame(
+    x_summary$coefficients
+    , row.names = NULL
+  )
+
+  res_table$Term <- rownames(x_summary$coefficients)
```
crsh, line 75: Don't we use `term` instead of `Term`?

R/apa_print_nlme.R
```diff
@@ -0,0 +1,133 @@
...
+#'
+#' @md
+#' @family apa_print
+#' @rdname apa_print.lme
+#' @method apa_print lme
+#' @export
+
+apa_print.lme <- function(
+  x
+  , args_confint
```
crsh, line 27: See above about `NULL`-default.

R/apa_print_nlme.R
```diff
@@ -0,0 +1,133 @@
...
+  res_table$conf.int <- unlist(
+    apply(X = confidence_intervals$fixed, MARGIN = 1, FUN = function(x) {
+      list(unname(x[c("lower", "upper")]))
+    })
+    , recursive = FALSE
+  )
+
+  attr(res_table$conf.int, "conf.level") <- args_confint$level
+
+  res_table$Term <- rownames(res_table)
```
crsh, line 81: See above about `term`.

R/apa_print_nlme.R
```diff
@@ -0,0 +1,133 @@
...
+
+  if(anyNA(x)) stop("Model-comparison tables of class 'anova.lme' are not supported.")
+
+  res_table <- as.data.frame(
+    x
+    , stringsAsFactors = FALSE
+    , make.names = TRUE
+  )
+
+  res_table$Term <- rownames(res_table)
```
crsh, line 118: See above about `term`.

R/apa_print_anova.R
```diff
@@ -137,34 +137,106 @@ apa_print.anova <- function(
...
+    if(any(grepl("Satterthwaite|Kenward", object_heading))) {
+      # determine correction type
+      sub_heading <- object_heading[grepl("Satterthwaite|Kenward", object_heading)][[1]]
+      attr(x, "correction") <- c("\\mathit{KR}", "S")[c(grepl("Kenward", sub_heading), grepl("Satterth", sub_heading))]
+
+      x$Effect <- rownames(x)
```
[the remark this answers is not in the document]
mariusbarth: It doesn't really matter, because I always set `rownames(x) <- NULL` within `beautify()` (so later in the chain). Maybe I should remove the call for `mixed` obejcts, too, to avoid redundancy.

R/apa_print_anova.R
mariusbarth: I think both cases are almost identical, but I didn't want to nest the code further before switching the other cases to the new pipeline. Thought it is more readable and a better starting point for creating a new "default" case later.

R/glue_apa_results.R
mariusbarth: Oh, sorry! Will remove this line.

R/apa_print_merMod.R
mariusbarth: No explicit reason, will change it to `NULL`

R/variable_label.R
mariusbarth: Hadley recommends using this construct because it is by magnitudes faster

R/utils.R
mariusbarth: Like that idea, will do.

R/utils.R
mariusbarth: done

R/utils.R
mariusbarth: Hm, true.

R/apa_print_merMod.R
mariusbarth: I do this because the lookup tables contain `Term = "term"`for column names  and `Term = "Term" for labels (similar to `Effect = "term"). So basically, here I mimic the input that `canonize()` is typically confronted with (instead of creating new cases in the lookup tables). The final result is a column with name `term` and label `Term`

R/apa_print_anova.R
```diff
@@ -137,34 +137,106 @@ apa_print.anova <- function(
...
+          , est_glue = construct_glue(beautiful_table, "estimate")
+          , stat_glue = construct_glue(beautiful_table, "statistic")
+          , term_names = sanitize_terms(x$Effect)
+          , in_paren = ellipsis$in_paren
+        )
+      )
+    }
+
+    # afex::mixed --------------------------------------------------------------
+    if(any(grepl("Mixed Model", object_heading))) {
```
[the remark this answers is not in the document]
mariusbarth: Hm, I would like to keep it like that, but that's probably something depending on personal taste/habits: I like to write it like that because the second `if` statement should (in this case) work independently of the first `if`... so basically, the order of if statements doesn't matter. I think using an explicit `else if` would falsely imply that we exploit the order of if statements. Again, I think this depends on your assumptions while reading code, and these may be quiet different.

R/apa_print_anova.R
```diff
@@ -137,34 +137,106 @@ apa_print.anova <- function(
...
-    # # lmerTest::ranova ----
-    # if(object_heading[1] == "ANOVA-like table for random-effects: Single term deletions") {
-    #   stop("Single-term deletions are not supported, yet.\nVisit https://github.com/crsh/papaja/issues to request support.")
-    # }
+
+    # lmerTest::anova.merModLmerTest -------------------------------------------
+    if(any(grepl("Satterthwaite|Kenward", object_heading))) {
+      # determine correction type
+      sub_heading <- object_heading[grepl("Satterthwaite|Kenward", object_heading)][[1]]
+      attr(x, "correction") <- c("\\mathit{KR}", "S")[c(grepl("Kenward", sub_heading), grepl("Satterth", sub_heading))]
```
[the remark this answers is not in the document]
mariusbarth: Good point, I will change that accordingly.

R/utils.R
mariusbarth: done

R/utils.R
mariusbarth: done

R/apa_print_merMod.R
mariusbarth: done

R/apa_print_nlme.R
mariusbarth: done

R/apa_print_anova.R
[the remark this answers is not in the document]
mariusbarth: done

R/glue_apa_results.R
mariusbarth: done

R/apa_print_anova.R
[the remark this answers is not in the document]
crsh: Well, the order is exploited by the serial processing of the script anyway. To me, `else if` makes clear that not both conditions can be met.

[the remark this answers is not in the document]
mariusbarth: done

mariusbarth:
> One thing I'm wondering: What about another table for random effects? I think it would be good to have it, even if there are no strings for reporting, yet, because adding them later will change the structure of the output (nested list in table).

I also wondered what might be the best approach here. In **broom.mixed**, `tidy.merMod` has an additional argument `effects` with which you can select whether fixed or random effects should be returned (and I think there is a similar `what` argument in `broom::tidy()`. Maybe this approach would also be good here, because this way the output of `apa_print()` can keep its standard structure. Moreover, it would ensure that our output objects don't become unwieldy when we implement even more complicated methods, like SEMs.

crsh:
I'm not sure, what you are proposing. Note that we already have at least one method that returns multiple tables. I don't see anything wrong with this. This is essentially how all of the other list elements are handled as well (no nesting if there is only one thing, think `htest`).

mariusbarth:
> Note that we already have at least one method that returns multiple tables.

Ah, I didn't know that. Which one?

Well, then `fixed_effects`?

crsh:
> Ah, I didn't know that. Which one?

`apa_print.list()` for `lm`s.

> Well, then `fixed_effects`?

:+1:

mariusbarth:
Am I missing something?
```
mod1 <- lm(Sepal.Length ~ Sepal.Width, data = iris)
mod2 <- update(mod1, formula = . ~ . + Petal.Length)
mod3 <- update(mod2, formula = . ~ . + Petal.Width)

# No bootstrapped Delta R^2 CI
model_comp <- apa_print(list(Baseline = mod1, Length = mod2, Both = mod3), boot_samples = 0)
str(model_comp)
List of 4
 $ estimate   :List of 2
  ..$ Length: chr "$\\Delta R^2 = .83$"
  ..$ Both  : chr "$\\Delta R^2 = .02$"
 $ statistic  :List of 2
  ..$ Length: chr "$F(1, 148) = 853.31$, $p < .001$"
  ..$ Both  : chr "$F(1, 148) = 19.04$, $p < .001$"
 $ full_result:List of 2
  ..$ Length: chr "$\\Delta R^2 = .83$, $F(1, 148) = 853.31$, $p < .001$"
  ..$ Both  : chr "$\\Delta R^2 = .02$, $F(1, 148) = 19.04$, $p < .001$"
 $ table      :Classes ‘apa_results_table’ and 'data.frame':	18 obs. of  3 variables:
  ..$ Baseline: chr [1:18] "6.53 [5.58, 7.47]" "-0.22 [-0.53, 0.08]" "" "" ...
  ..$ Length  : chr [1:18] "2.25 [1.76, 2.74]" "0.60 [0.46, 0.73]" "0.47 [0.44, 0.51]" "" ...
  ..$ Both    : chr [1:18] "1.86 [1.36, 2.35]" "0.65 [0.52, 0.78]" "0.71 [0.60, 0.82]" "-0.56 [-0.81, -0.30]" ...
 - attr(*, "class")= chr [1:2] "apa_results" "list"
```

crsh:
Hmm, no I must be misremembering and am too tired to read my code. ;)

But what's the alternative here?

mariusbarth:
Haha, no worries!

[Here](https://www.tidymodels.org/learn/develop/broom/) it is described what I mean:

>Some things to keep in mind while writing your tidy() method:
> - Sometimes a model will have several different types of components. For example, in mixed models, there is different information associated with fixed effects and random effects. Since this information doesn’t have the same interpretation, it doesn’t make sense to summarize the fixed and random effects in the same table. In cases like this you should add an argument that allows the user to specify which type of information they want. For example, you might implement an interface along the lines of:

```
model <- mixed_model(...)
tidy(model, effects = "fixed")
tidy(model, effects = "random")
```


I'm also not sure what solution is better (combining all information in one big output list vs. making the output conditional on a further argument). One more thing that we will have to think about if we decide for one big output is how to include reporting strings (i.e., `statistic`, `estimate`, `full_result`) for random effects in the future. But I'm more and more inclined towards one big output. If, in the future, we consider the output to become unwieldy for complicated models, we could still write methods to select information, for instance:
```
complicated_model %>%
  apa_print() %>%
  random_effects()
```

crsh:
Ah, now we are on the same page. You are basically proposing to extend the `tidy()`-argument to `apa_print()`. Given that this is an important design question, we may want to invite others to the discussion. I'll have initiated a discussion about this in the [parent issue](https://github.com/crsh/papaja/issues/154#issuecomment-670401115).

crsh:
Okay, we got a few responses and the conditional table  seems to be preferred by us and all users who responded. As far as I understand, for now this just means adding the argument but only accepting `effects = "fixed"` and documenting that random effects will follow. If you could add this minor change I'll do a last round of testing and most likely just accept the PR.

mariusbarth:
> If you could add this minor change I'll do a last round of testing and most likely just accept the PR.

Done.

R/apa_print_nlme.R
```diff
@@ -0,0 +1,133 @@
...
+#' and create formatted character strings report the results in accordance
+#' with APA manuscript guidelines.
+#'
+#' @param x A (non-)linear mixed-effects model fitted with [nlme::lme()] or [nlme::nlme()]
+#' @param args_confint Named list. Arguments that are passed to [nlme::intervals.lme()].
+#' @inheritParams apa_print.merMod
+#' @examples
+#'   library(nlme)
+#'   fm1 <- lme(distance ~ age, data = Orthodont, method = "ML") # random is ~ age
+#'   fm2 <- lme(distance ~ age + Sex, data = Orthodont, random = ~ 1, method = "ML")
```
crsh, line 13: You fit three models here, but only the first one ever gets used in the example. Maybe drop the two others?

R/apa_print_merMod.R
```diff
@@ -1,144 +1,119 @@
...
+#'   (`"fixed"`) are supported.
+#' @param args_confint Named list. Additional arguments that are passed to [lme4::confint.merMod()], see details.
+#' @param est_name     An optional character. The label to be used for fixed-effects coefficients.
+#' @inheritParams print_anova
+#' @param ... Further arguments that are passed to [printnum()].
+#' @details
+#'   Confidence intervals are calculated by calling [lme4::confint.merMod()]. By
+#'   default, *Wald* confidence intervals are calculated, but this may change in
+#'   the future.
+#'
```
crsh, line 19: Could you also add a brief example here?

crsh (changes requested):
Looks good, thanks. Just the minor points about the examples and I have another petty thing for which I'll open a separate issue.

mariusbarth:
> Looks good, thanks. Just the minor points about the examples and I have another petty thing for which I'll open a separate issue.

I addressed the issues concerning the examples. I also changed a minor point: The new default for printing a progress bar is the value returned by `interactive()`.

crsh:
Great work.

Issue #364, opened 2020-05-11
How to unitize `apa_results_table`


mariusbarth:
Hi @crsh, while working on LMM support, I (again) started wondering how we could best unitize the output of `apa_print()`. If I remember correctly, we both agreed that it would be desirable to have a column `statistic` irrespective of the actual type of test statistic (F, t, chi^2 value) in an `apa_results_table`; however, we also did not want to create a code-breaking change. I think I came up with a possible solution. We could create a new method for indexing objects of class `apa_results_table`:

```
#' @export

`$.apa_results_table` <- function(x, name) {
  if(name == "F") {
    message("Indexing an apa_results_table with `$F` is deprecated. Use `$statistic` instead.")
    return(x$statistic)
  } else {
    x[[name]]
  }
}
```

With a method like this, we could rename the respective column in all ANOVA outputs. What do you think about this solution?

crsh:
Hmm, I have considered this, too, but thus far it has struck me as a "terrible hack". ;)

But I'm coming around to the idea. My initial hesitation was that this only works for element selection but might break in unexpected ways when other functions are applied to the list. But then again, I'm not sure how often this would be the case and if it's worth to keep the inconsistencies for the sake of full backward compatibility here.

So, let's give it a shot. Off the top of my head, two things we need to think about:

1. I think we would need additional methods for  `[[` and `[`.
2. To avoid more hacks of this we need to think hard about what column names we have and may need in the future and how they can be unified (I'm thinking about things like  `terms`, `predictors`, `contrasts` etc.). Also, I would suggest that we follow the naming conventions used in `broom` where they exist.

What do you think?

mariusbarth:
Great, I totally agree! Maybe I will just start collecting my ideas.

What needs to be done:
- Rename columns `F`, `t`, etc to `statistic`
- Rename column `p` to `p.value` (for consistency with broom and most summary methods)
- Rename columns Effect, Predict, etc to `term` (because it's the most generic form)
- I think we should stick with `estimate` for original-scale estimates.

What might be hard:
- Sometimes we have multiple test statistics in one row, possibly with extra "parameter" (e.g., for MANOVA both multivariate and univariate test statistic, both with their own df). I'm leaning towards prepending the respective terms in these case (so in this case, `multivariate.statistic`, `multivariate.df`). This is inconsistent with broom, but I think I should fix my own mistake, there.
[...]

PR #415, opened 2020-10-08
Standardize apa_results_table

mariusbarth:
This PR includes standardization for `apa_results_table` outputs.

It also includes extraction methods for these tables, so that we soft-deprecate the old column names (a warning is given). Check out `test_apa_results_table.R` for examples.

- rename column `ci` to `conf.int`
- rename columns `predictor` and `effect` to `term`
- rename `F` and `chisq` to `statistic`
- rename column `es`, `pes`, `ges` to `estimate`
- only return one `estimate` for ANOVA, warn if length > 1
- Levene test: remove nesting structure
- bugfix in apa_print.papaja_wsci: inherit dot arguments


R/apa_results_table.R
```diff
@@ -54,32 +54,128 @@ print.apa_results_table <- function(x, ...) {
...
+    , "ges"       = "estimate"
+    , "es"        = "estimate"
+    , "ci"        = "conf.int"
+    , "predictor" = "term"
+  )
+  new_name <- aliases[pmatch(name, names(aliases), nomatch = 0L)]
+
+  if(length(new_name) > 0L) {
+    warning(
+      "Indexing an apa_results_table with '$"
```
crsh, line 92: How do you feel about making this a little more explanatory? Suggestion: `"To improve consistency of apa_print()-output, the column '", name, "' has been renamed to '", unname(new_name), "'. The desired values were returned, but please update your code accordingly, as we will drop support for the old column names in a future release."`

crsh:
Excellent work! This should resolve https://github.com/crsh/papaja/issues/364. I only left a minor comment regarding the aliased index warning.

R/apa_results_table.R
mariusbarth: Done

PR #420, opened 2020-10-18
Snapshot tests

mariusbarth:
Here I collected some changes that increase line coverage of our tests. Also, CI should also run through now.


crsh (approved):
Great, thanks!

PR #429, opened 2020-11-03
set default csl arguments for revision letter

conig:
The revision letter currently cannot be knitted due to changes to set_default_csl.
This request sets those arguments to restore functionality.


crsh:
Thanks!

Issue #322, opened 2019-10-23
Could not create a copy of the original input file xxx while trying to render the appendix.


bnicenboim:
I was getting the `Could not create a copy of the original input file **filename** while trying to render the appendix.` when I added an appendix.

I already know that this is because of line 717 in `https://github.com/crsh/papaja/blob/master/R/apa6_formats.R` The problem was that a temporary file was created and not deleted automatically. I fixed this by deleting the file.

I just wanted to know if I should change the code and do a pull request: I could add either a more informative message that tells the user to manually delete de temporary file, or make papaja do it by itself.

crsh:
That's a good point. I thought I had done this already. I'll try adding this to `on_exit`.

crsh:
Hi @karchjd, thanks for the example and sorry for the delay. I'm sorry to report that I cannot reproduce the problem with the files you attached to your last post; the original `paper.Rmd` is restored and `cGFwZXIuUm1k.Rmd`is deleted automatically. Just to be sure, could you retry after reinstalling the latest development version?

~~~r
remotes::install_github("crsh/papaja@devel")
~~~

PR #432, opened 2020-12-02
Correctly clean temporary file if appendix is used

mutlusun:
Hello,

having `appendix: "appendix.Rmd"` in my YAML-Header, the temporary input file was not cleaned up. This resulted in the following error on the following run (cf. #322 ):

```
Error in modify_input_file(input = input, format = "papaja::apa6_docx") : 
  Could not create a copy of the original input file 'XXX/main.Rmd' while trying to render the appendix.
```

The temporary input file was not cleaned up, because `rmarkdown::metadata$appendix` is not available on exit. Thus, the YAML header has to be reread on exit.

### DIFF
--- a/R/apa6_formats.R
+++ b/R/apa6_formats.R
@@ -714,7 +714,9 @@ revert_original_input_file <- function(x = 1) {
   input_file <- get("original_input", envir = parent.frame(x))
   input_file <- tools::file_path_as_absolute(input_file)
 
-  if(!is.null(rmarkdown::metadata$appendix)) {
+  yaml_params <- get_yaml_params(readLines(input_file))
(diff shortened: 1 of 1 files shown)


crsh:
I‘m on the go, but wanted to give quick feedback: I think this issue may already be fixed on the `devel` branch. Could you check?

mutlusun:
You are right. This bug has already be fixed in the devel branch. But I think your implementation is a more clever solution ;) Do you mind merging this into master? I have cherry-picked your patch.

crsh:
Absolutely, thanks again for reporting the problem and contributing a fix. Very much appreciated.

PR #434, opened 2020-12-10
Report correct denominator df for comparisons of linear models

### LINKED ISSUE #433: apa_print.list reports wrong error-df for lm model comparison?
Thanks for this very helpful package! Just a small potential bug: As I understand it, model comparisons should report the error degrees of freedom from the test of the difference (i.e. the bottom row in the `anova()` output. Currently, apa_print.list reports the error df from the top row, which I think is incorrect? Shouldn't it be F(1,28) in the example below?

``` r
mod1 <- lm(mpg ~ wt + cyl, mtcars)
mod2 <- lm(mpg ~ wt * cyl, mtcars)
anova(mod1, mod2)
#> Analysis of Variance Table
#> 
#> Model 1: mpg ~ wt + cyl
#> Model 2: mpg ~ wt * cyl
#>   Res.Df    RSS Df Sum of Sq      F  Pr(>F)  
#> 1     29 191.17                              
#> 2     28 156.98  1    34.196 6.0995 0.01988 *
#> ---
#> Signif. codes:  0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1
papaja::apa_print(list(mod1, mod2), boot = 0)$statistic
#> $model2
#> [1] "$F(1, 29) = 6.10$, $p = .020$"
```

<sup>Created on 2020-12-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

mariusbarth:
This is a hotfix that fixes #433. Denominator degrees of freedom for differences in R² were always taken from the baseline model, but should be taken from the more-complex model. (Note that *p* values were still reported for the **correct** degrees of freedom, because we extracted these from the output of `anova.lmlist()`.)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,6 +53,6 @@ SystemRequirements: Rendering the document template requires pandoc (>= 2.0; htt
 License: MIT + file LICENSE
 LazyData: yes
 Encoding: UTF-8
-RoxygenNote: 7.1.0.9000
+RoxygenNote: 7.1.1
 VignetteBuilder: knitr
 Language: en-US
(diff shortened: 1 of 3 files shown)


crsh (approved):
Thanks for the quick fix.

Issue #435, opened 2020-12-22
Knitting the papaja::revision_letter skeleton fails with a CSLReferences Undefined Error


briandk:
First, THANK YOU for making papaja. It has made my life so much easier!

**Describe the bug**
Knitting the  `papaja::revision_letter` skeleton fails with a `CSLReferences Undefined` error in LaTeX.

**To Reproduce**
1. Launch RStudio
2. File -> New File -> RMarkdwon...
3. From Template -> Revision Letter (PDF) `{papaja}`
4. Knit -> Knit to revision_letter_pdf
5. Save the .Rmd file
6. Observe the following output in the RMarkdown Console:

```
processing file: Revision-Letter-2020-12-22.Rmd
  |......................................................................| 100%
   inline R code fragments


/Applications/RStudio.app/Contents/MacOS/pandoc/pandoc +RTS -K512m -RTS Revision-Letter-2020-12-22.utf8.md --to latex --from markdown+autolink_bare_uris+tex_math_single_backslash --output Revision-Letter-2020-12-22.tex --lua-filter /Library/Frameworks/R.framework/Versions/4.0/Resources/library/bookdown/rmarkdown/lua/custom-environment.lua --lua-filter /Library/Frameworks/R.framework/Versions/4.0/Resources/library/rmarkdown/rmarkdown/lua/pagebreak.lua --lua-filter /Library/Frameworks/R.framework/Versions/4.0/Resources/library/rmarkdown/rmarkdown/lua/latex-div.lua --self-contained --table-of-contents --toc-depth 2 --template /Library/Frameworks/R.framework/Versions/4.0/Resources/library/papaja/rmarkdown/templates/revision_letter/resources/revision_letter.tex --number-sections --highlight-style tango --pdf-engine pdflatex --wrap preserve --csl /Library/Frameworks/R.framework/Versions/4.0/Resources/library/papaja/rmd/apa6.csl --citeproc --lua-filter /Library/Frameworks/R.framework/Versions/4.0/Resources/library/rmdfiltr/replace_ampersands.lua --lua-filter /Library/Frameworks/R.framework/Versions/4.0/Resources/library/rmdfiltr/wordcount.lua 
output file: Revision-Letter-2020-12-22.knit.md

67 words in text body
0 words in reference section
! LaTeX Error: Environment CSLReferences undefined.

[...]

PR #438, opened 2021-02-15
patch for CSLReferences error

conig:
As reported in issue #435 papaja::revision_letter is currently failing due to an update to pandoc's definition of the cslreferences environment.

@crsh suggested changes to the latex skeleton in order to fix this issue.

This pull request changes the revise and resubmit LaTeX and rmd skeletons in order to integrate these fixes.


crsh (approved):
Thanks a lot. Not sure why this was sitting idle for so long. Much appreciated!

PR #443, opened 2021-03-15
Add custom effect sizes

mariusbarth:
Hi @crsh,

I finished work on custom effect sizes for ANOVA. Take a look at it! Unfortunately, few outputs from `summary()` (in fact, only `summary.Anova.mlm`) will not be supported, but with the rest everything went very well even with very standard programming techniques (i.e., without fiddling with environments).

Some things to keep in mind and possibly discuss:

- There are now three different ways to include standardized effect-size measures:
  a) Pass one of `"ges"`, `"pes"`, or `"es"`, and our good old effect-size function will do the rest
  b) Pass a function like `effectsize::eta_squared()`
  c) Pass a data frame with appropriate columns
- When using b), I opted against passing an extra list of arguments to the effect-size function, because I think if a user wishes to have that level of control, they can choose the "data-frame method" c) and first execute the custom function with all necessary arguments, and then pass the result to `apa_print()`.
- I think my preferred choice of using the new functionality would look like the following:
```
# Define my custom effect-size function
ges <- function(x, ...) {
  effectsize::eta_squared(x, generalized = TRUE, include_intercept = TRUE, ...)
}
# And then use it for concise code
anova_out <- afex::aov_4(formula = yield ~ N * P, npk)
apa_print(anova_out, es = ges)
```
- `include_intercept = TRUE` ensures that the effect size for the intercept term is available if I want to include it (If the user forgets to require the intercept, they will get an informative error message)


crsh:
Hey Marius, thanks a lot for getting started on this. I've started reviewing, but I need some more time to think this big change through. But maybe we can start by discussion your above points:

1. Could you try to give me a rough idea as to why `summary.Anova.mlm` cannot be supported?
2. Preferably, I would retire using our code for a) where possible because it does not yield any confidence intervals. At least in case `effectsize` is available, I would automatically apply the corresponding function from the package when a string is passed. What do you think?
3. As you point out, besides the felxible but less convenient`data.frame`-option, it is always possible to define custom functions to "pass arguments" to a function. While this seems straight forward to me (and also has high reuse potential across analyses in any given manuscript) I do think it is less accessible to many users. Would it be a problem to add the argument-list as additional option?

mariusbarth:
Hi Frederik,

1. The problem with `summary.Anova.mlm` is that these objects are not supported by `effectsize` **and** do not contain a call that would enable us to re-create the corresponding `Anova.mlm` object. I could not come up with a solution to recover the lost information.
2. I think in the long run it is preferable to include confidence intervals as a default, but I think we should be conservative here and stick with our effect-size function for two reasons:
    1. If we use `effectsize` as a default if it is installed, the output of a papaja document will depend on whether this package is installed or not on the specific machine. With reproducibility in mind, I really don't like the idea.
    2. It is my impression that using the `effectsize` package as a "proper" default (i.e., importing it and setting the respective defaults) is a bit premature, given its lifecycle.
3. It is possible to add an additional argument (e.g., `args_estimate`), but while I was considering this option, I figured it would not really help in writing concise code. For instance, with a very simple example:

```
# data-frame method
apa_print(anova_out, estimate = eta_squared(anova_out, generalized = TRUE))

# A method with an arguments list
apa_print(anova_out, estimate = eta_squared, args_estimate = list(generalized = TRUE))

So I thought we should better encourage using the data-frame method. But I don't have a strong opinion on this last one.

**Comment by crsh** [comment]
> If we use effectsize as a default if it is installed, the output of a papaja document will depend on whether this package is installed or not on the specific machine. With reproducibility in mind, I really don't like the idea.

I'm pretty torn on this. On the one hand, it's a fair point; we wouldn't want manuscripts that are suddenly missing confidence intervals (although we could produce a message here, especially because we may consider deprecating the character option). On the other hand, the headaches that package updates have caused me over the years have convinced me that computational reproducibility is too fragile without conservation of the computational environment (i.e. software). This is especially true in an ecosystem that is heavily modular with large dependency trees and that, in large parts, is maintained by self-taught developers (ourselves included) and developlers who favor maintainability and progress over backward compatibility and stability (very understandable, especially when most everyone works for free).

> I figured it would not really help in writing concise code

I see, that's true. It would, however, allow users to prespecify a list of arguments that can be reused across `apa_print()`-calls (much like writing a custom function, but potentially a less daunting?). Not sure, whether it's worth "complicating" the interface, risking bugs, and putting in additional effort...

**Comment by mariusbarth** [comment]
After giving it some more thought, I believe we should indeed pursue the long-term goal to use the **effectsize** package as a default, and yes, we should make this conditional on whether it is installed or not. I am hesitant to using it as a default right now, because it is my impression that the project is a bit too young, and the user interface and default behavior might be subject to too many changes. To make it possible for users to customize effect-size measures and ensure consistent reporting, I would opt into the approach using options (see above) to define an effect-size function within `.onLoad()`.

Within `.onLoad()`, we could implement it along the lines of this:

```
if(package_available("effectsize")) {
  op_papaja$papaja.estimate_anova <- function(x, generalized = TRUE, ...) {
    effectsize::eta_squared(x, generalized = generalized, ...)
  }
}
```

What do you think?

**Comment by crsh** [comment]
I think this sounds good to me. So do you envision the default arguments to look like this?

~~~r
apa_print(x, es = getOption("papaja.estimate_anova"))
~~~

And extend the `.onLoad()` as follows to retain the current behavior?

~~~r
if(package_available("effectsize")) {
  op_papaja$papaja.estimate_anova <- function(x, generalized = TRUE, ...) {
    effectsize::eta_squared(x, generalized = generalized, ...)
  }
} else {
  op_papaja$papaja.estimate_anova <- "ges"
}

~~~

**Comment by mariusbarth** [comment]
Yes, this looks perfect to me. If you want, I can add the above lines of code to the branch.

**Comment by crsh** [comment]
That would be great.

**Comment by mariusbarth** [comment]
I just committed the above-mentioned changes, with the following exceptions:

- If the **effectsize** package is not available, the default is set to "ges" -- however, I don't do this while loading the package, but within the function definition via `getOption("papaja.estimate_anova", default = "ges"). This ensures that the default is available (and not `NULL`) when `apa_print()` is used without loading **papaja**.
- I also instruct `effectsize::eta_squared()` to include intercepts by default, which ensures that it is always available.

**Comment by crsh** [comment]
Great. Just one more question: If we include the intercept by default (rather than depending on user choices) wouldn't this lead to different results than not including it for partial and generalized effect size estimates?

File: R/apa_print_anova.R [hunk 1]
```diff
@@ -61,66 +73,180 @@ apa_print.summary.aov <- function(x, ...) {
...
+  canonical_table <- add_custom_effect_sizes(
+    canonical_table = canonical_table
+    , .x = .x
+    , estimate = estimate
+    , mse = mse
+    , observed = observed
+    , intercept = intercept
+  )
+
+  if(!intercept) canonical_table <- canonical_table[canonical_table$term != "(Intercept)", , drop = FALSE]
```
crsh, line 120: Doesn't this already happen in `add_custom_effect_size()`?

R/apa_print_anova.R
```diff
@@ -61,66 +73,180 @@ apa_print.summary.aov <- function(x, ...) {
...
+    canonical_table
+    , estimate = estimate
+    , mse = mse
+    , observed = observed
+    , intercept = intercept
+    , .x = .x
+  )
+
+  # Remove intercept if the user doesn't want it:
+  if(!intercept) canonical_table <- canonical_table[canonical_table$term != "(Intercept)", , drop = FALSE]
```
crsh, line 201: Doesn't this already happen in `add_custom_effect_size()`?

R/add_custom_effect_sizes.R
```diff
@@ -0,0 +1,39 @@
...
+
+  if(!intercept) canonical_table <- canonical_table[canonical_table$term != "(Intercept)", , drop = FALSE]
+
+  y <- merge(canonical_table,
+             tidy_es(estimate)
+             , sort = FALSE
+             , all.x = TRUE # Do not drop
+  )
+  if(anyNA(y)) {
+    warning("Custom effect sizes were not available for some model terms. These have been dropped from the output object.", call. = FALSE)
```
crsh, line 27: Hmm, when does this typically happen? And wouldn't it make more sense to retain the term but just omit the effect size?

tests/testthat/test_print_anova.R
```diff
@@ -13,7 +13,7 @@ test_that(
     tw_aov1 <- apa_print(aov(yield ~ N * P, npk), es = c("ges"), observed = "N")
     tw_aov2 <- apa_print(aov(yield ~ N * P, npk), es = c("pes"), observed = "N")
     tw_aov3 <- apa_print(aov(yield ~ N * P, npk), es = c("es"), observed = "N")
-    expect_equal(tw_aov1$estimate$N, "$\\hat{\\eta}^2_G = .224$")
+    expect_equal(tw_aov1$estimate$N, "$\\hat{\\eta}^2_G = .218$")
```
crsh, line 16: Hmm, why has this value changed?

crsh (commented):
Looks great. Once the comments have been addressed I'll take one more look and take it for a spin.

R/apa_print_anova.R
```diff
@@ -4,17 +4,29 @@
...
 #'    of freedom, are replaced with brackets.
 #'
+#'    Argument `estimate` determines which measure of effect size is to be used: It is currently possible to provide
+#'    a character, where `"ges"` calculates generalized eta squared,`"pes"` calculates partial eta squared, and `"es"`
+#'    calculates eta squared. Note that eta squared is calculated correctly if and only if the design is balanced.
+#'
+#'    It is also possible to provide a `data frame` with columns `estimate`, `conf.low`, and `conf.high`, which allows
+#'    for including custom effect-size measures.
+#'
+#'    A third option is to provide a function that will be used to calculate effect-size measures from `x`. See the
```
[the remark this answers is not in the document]
mariusbarth: What do you mean in addition to "with columns `estimate`, `conf.low`, and `conf.high`"?

R/apa_print_anova.R
```diff
@@ -61,66 +73,180 @@ apa_print.summary.aov <- function(x, ...) {
...
+  , estimate = getOption("papaja.estimate_anova", "ges")
+  , mse = TRUE
+  , observed = NULL
+  , intercept = FALSE
+  , in_paren = FALSE
+  , ...
+) {
+
+  ellipsis <- list(...)
+  .x <- ellipsis$.x
```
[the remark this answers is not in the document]
mariusbarth: Against reconstruction: A `summary.aovlist` does not contain the call.
```
> str(summary(afex::aov_4(yield ~ (N|block), npk)$aov))
List of 2
 $ Error: block  :List of 1
  ..$ :Classes ‘anova’ and 'data.frame':	1 obs. of  5 variables:
  .. ..$ Df     : num 5
  .. ..$ Sum Sq : num 172
  .. ..$ Mean Sq: num 34.3
  .. ..$ F value: num NA
  .. ..$ Pr(>F) : num NA
  ..- attr(*, "class")= chr [1:2] "summary.aov" "listof"
 $ Error: block:N:List of 1
  ..$ :Classes ‘anova’ and 'data.frame':	2 obs. of  5 variables:
  .. ..$ Df     : num [1:2] 1 5
  .. ..$ Sum Sq : num [1:2] 94.6 49.3
  .. ..$ Mean Sq: num [1:2] 94.64 9.85
  .. ..$ F value: num [1:2] 9.61 NA
  .. ..$ Pr(>F) : num [1:2] 0.0269 NA
  ..- attr(*, "class")= chr [1:2] "summary.aov" "listof"
 - attr(*, "class")= chr "summary.aovlist"

```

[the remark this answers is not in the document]
mariusbarth: Output looks like the following:
```
> library(papaja)
Lade nötiges Paket: tinylabels
> apa_print(summary(afex::aov_4(yield ~ (N|block), npk)$aov))
Fehler: Cannot apply custom effect-size function to this class of object.
Zusätzlich: Warnmeldung:
 Fehler: Cannot apply custom effect-size function to this class of object. 
> apa_print(summary(afex::aov_4(yield ~ (N|block), npk)$aov), estimate = "ges")
$estimate
$estimate$N
[1] "$\\hat{\\eta}^2_G = .300$"


$statistic
$statistic$N
[1] "$F(1, 5) = 9.61$, $\\mathit{MSE} = 9.85$, $p = .027$"


$full_result
$full_result$N
[1] "$F(1, 5) = 9.61$, $\\mathit{MSE} = 9.85$, $p = .027$, $\\hat{\\eta}^2_G = .300$"


$table
A data.frame with 7 labelled columns:

  term estimate statistic df df.residual  mse p.value
1    N     .300      9.61  1           5 9.85    .027

term       : Effect 
estimate   : $\\hat{\\eta}^2_G$ 
statistic  : $F$ 
df         : $\\mathit{df}$ 
df.residual: $\\mathit{df}_{\\mathrm{res}}$ 
... (2 more labels)
attr(,"class")
[1] "apa_results" "list"       
Warnmeldung:
More than one observation per cell, aggregating the data using mean (i.e, fun_aggregate = mean)! 
```

tests/testthat/test_print_anova.R
mariusbarth: Ìf I remember correctly, `observed` has not been passed properly before.

R/add_custom_effect_sizes.R
mariusbarth: If the user chooses an effect-size function that does not calculate effect sized for intercept terms, this will happen. I think it is better to warn here, because this will most likely be a mistake. But I will remove the dropping-terms part, maybe that's a bit harsh.

R/apa_print_anova.R
mariusbarth: Tbh, dunno. There were so many redundancies in the ANOVA code that I thought it might be too far-fetched to remove all of them. Think this PR is a step into the right direction, but more code refactoring will be necessary in the future. But please, let's not mix up this new feature with a refactoring of the ANOVA code, because then it will be implemented in 12 years

R/apa_print_anova.R
mariusbarth: see below

R/add_custom_effect_sizes.R
```diff
@@ -0,0 +1,39 @@
+
+
+add_custom_effect_sizes <- function(estimate, ...) {
+  UseMethod("add_custom_effect_sizes", estimate)
+}
+
+
+add_custom_effect_sizes.character <- function(estimate, canonical_table, .x = NULL, ...) {
+
+  if(length(estimate) > 1L) warning("Calculating more than one effect-size measure is now deprecated.")
```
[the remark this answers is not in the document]
mariusbarth: ok

R/apa_print_anova.R
```diff
@@ -61,66 +73,180 @@ apa_print.summary.aov <- function(x, ...) {
...
 }
 
 
 #' @rdname apa_print.aov
 #' @method apa_print summary.Anova.mlm
 #' @export
 
-apa_print.summary.Anova.mlm <- function(x, correction = getOption("papaja.sphericity_correction"), intercept = FALSE, ...) {
-  variance_table <- arrange_anova(x, correction)
+apa_print.summary.Anova.mlm <- function(
```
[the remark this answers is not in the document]
mariusbarth: see below

tests/testthat/test_print_anova.R
mariusbarth: Think I remember correctly:
```
> afex::aov_4(yield ~ (N*P|block), npk)
Anova Table (Type 3 tests)

Response: yield
  Effect   df   MSE      F  ges p.value
1      N 1, 5 19.70 9.61 * .224    .027
2      P 1, 5 14.28   0.59 .013    .478
3    N:P 1, 5 28.84   0.74 .031    .430
---
Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘+’ 0.1 ‘ ’ 1
> afex::aov_4(yield ~ (N*P|block), npk, observed = "N")
Anova Table (Type 3 tests)

Response: yield
  Effect   df   MSE      F  ges p.value
1      N 1, 5 19.70 9.61 * .218    .027
2      P 1, 5 14.28   0.59 .010    .478
3    N:P 1, 5 28.84   0.74 .025    .430
---
Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘+’ 0.1 ‘ ’ 1
```

R/apa_print_anova.R
mariusbarth: Ok, now I'm hooked on getting better code: I think I should remove the code from `add_custom_effect_size()` because it really does not really belong there.

R/apa_print_anova.R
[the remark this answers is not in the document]
crsh: I'm referring to the third option here.

R/add_custom_effect_sizes.R
mariusbarth: If I leave the NA, there are some problems that need to be addressed downstream: If we allow NAs in an `apa_results_table`, how should they be included? Should it be `"NA"` or `NA_character_`? I'm in favor of `NA_character_`, but this will need to be addressed within `beautify` and possibly `glue_apa_results()`.

crsh: `printnum()` has an argument, which draws on a global option to process this, no? But if it's too much of a hassle, just keep it as is!

tests/testthat/test_print_anova.R
crsh: Hmm, that's not so great...

R/add_custom_effect_sizes.R
mariusbarth: OK, after playing around with this one I determined it is best actually the best solution to drop the terms with NAs. Keeping terms with NAs, I had to switch off lots of input validation that is happening down the road, and we would have to write lots of code handling these rare (and most likely non-intentional) cases.

R/apa_print_anova.R
mariusbarth: I also played around with this one. Basically, removing these lines of code leads to side effects with our internal effect-size function. We should open a new issue that aims at really refactoring this function (i.e., making an effect-size function such as `effectsize::eta_squared()`, then we will be able to substantially simplify the ANOVA code.

R/apa_print_anova.R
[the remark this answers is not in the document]
mariusbarth: We only support **effectsize** right now, so I would keep it like that. I would not encourage users to write their own functions that we call from within `apa_print()`, because then we would have to maintain tidiers for these functions. If a user wants to write their own function to calculate effect sizes, their function should most likely just create an output that is understood by the data-frame method.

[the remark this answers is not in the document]
crsh: Okay, so should we change this to "a function from the effectsize package" to make it crystal clear?

R/apa_print_anova.R
crsh: Agreed. Given that you have a better grasp on what's required here, could you open an issue for this?

R/apa_print_anova.R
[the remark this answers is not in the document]
mariusbarth: done

R/apa_print_anova.R
mariusbarth: done

mariusbarth:
> Great. Just one more question: If we include the intercept by default (rather than depending on user choices) wouldn't this lead to different results than not including it for partial and generalized effect size estimates?

It should only matter for calculating (standard) eta squared. However, look at this:
```
> anova_out <- afex::aov_4(yield ~ (N*P|block), npk)
> effectsize::eta_squared(anova_out, partial = FALSE, include_intercept = TRUE)
# Effect Size for ANOVA (Type III)

Parameter   |     Eta2 |       90% CI
-------------------------------------
(Intercept) |    82.47 | [0.00, 0.00]
N           |     0.22 | [0.00, 0.62]
P           | 9.59e-03 | [0.00, 0.33]
N:P         |     0.02 | [0.00, 0.41]
```
So I think we should open an issue at https://github.com/easystats/effectsize/issues.

crsh:
Okay, I guess that makes sense. There is another issue with the default settings (and their interaction with `effectsize`), though. I'll quote their documentation:

> `generalized`: If `TRUE`, returns generalized Eta Squared, assuming all variables are manipulated. Can also be a character vector of observed (non-manipulated) variables, in which case generalized Eta Squared is calculated taking these observed variables into account. For afex_aov model, when `generalized = TRUE`, the observed variables are extracted automatically from the fitted model, if they were provided then.

If we default to `TRUE` this will cause inconsistent results for equivalent models depending on the function that was used to fit the model/on the object type.

~~~r
library("afex")
library("effectsize")
data(obk.long)

afex_aov_between <- aov_ez(
    "id"
    , "value"
    , obk.long
    , between = c("treatment", "gender")
    , observed = c("gender")
)

eta_squared(
    afex_aov_between
    , generalized = TRUE
)
~~~
~~~
Parameter        | Eta2 (generalized) |       90% CI
----------------------------------------------------
treatment        |               0.29 | [0.00, 0.56]
gender           |               0.19 | [0.00, 0.51]
treatment:gender |               0.29 | [0.00, 0.57]

- Observed variabels: gender
~~~
~~~r
eta_squared(
    afex_aov_between$Anova
    , generalized = TRUE
)
~~~
~~~
Parameter        | Eta2 (generalized) |       90% CI
----------------------------------------------------
treatment        |               0.44 | [0.00, 0.67]
gender           |               0.27 | [0.00, 0.57]
treatment:gender |               0.36 | [0.00, 0.62]

- Observed variabels: All
~~~
~~~r
eta_squared(
    afex_aov_between$Anova
    , generalized = "gender"
)
~~~
~~~
Parameter        | Eta2 (generalized) |       90% CI
----------------------------------------------------
treatment        |               0.29 | [0.00, 0.56]
gender           |               0.19 | [0.00, 0.51]
treatment:gender |               0.29 | [0.00, 0.57]

- Observed variabels: gender
~~~

I'm not sure how to deal with this other than asking the developers not to overload the `generalized` parameter. Relatedly, our `observed` argument in `apa_print()` is currently ignored, which I found rather confusing.

There is one more thing that I don't understand. I would expect the following three to give equivalent results, but they don't:

~~~r
afex_aov_between <- aov_ez(
    "id"
    , "value"
    , obk.long
    , between = c("treatment", "gender")
    , observed = c("treatment", "gender")
)

eta_squared(
    afex_aov_between
    , generalized = TRUE
)
~~~
~~~
Parameter        | Eta2 (generalized) |       90% CI
----------------------------------------------------
treatment        |               0.29 | [0.00, 0.56]
gender           |               0.13 | [0.00, 0.46]
treatment:gender |               0.21 | [0.00, 0.49]

- Observed variabels: treatment, gender
~~~
~~~r 
eta_squared(
    afex_aov_between$Anova
    , generalized = c("treatment", "gender")
)
~~~
~~~
Parameter        | Eta2 (generalized) |       90% CI
----------------------------------------------------
treatment        |               0.29 | [0.00, 0.56]
gender           |               0.13 | [0.00, 0.46]
treatment:gender |               0.21 | [0.00, 0.49]

- Observed variabels: treatment, gender
~~~
~~~r 
eta_squared(
    afex_aov_between$Anova
    , generalized = TRUE
)
~~~
~~~
Parameter        | Eta2 (generalized) |       90% CI
----------------------------------------------------
treatment        |               0.44 | [0.00, 0.67]
gender           |               0.27 | [0.00, 0.57]
treatment:gender |               0.36 | [0.00, 0.62]

- Observed variabels: All
~~~

Am I missing something obvious here?

mariusbarth:
Oh wow, my fault! I didn't read the documentation properly. So we will need to add `generalized = observed` or something equivalent to the default effect-size function.

crsh:
Sorry, I revised my previous comment. I totally misread the documentation. :/

mariusbarth:
I don't think you are missing something here, but that this is a bug.

Still, we will have to add the necessary information about observed vs. manipulated factors to our default effect-size function. Maybe (within `apa_print()`), we add the necessary information as an attribute and read this out from within the effect-size function (the same way as it already does for `afex_aov` objects),

mariusbarth:
You are right, it should read `"Observed variables: None"`.

mariusbarth:
I added a draft for passing `observed` to the custom function

crsh:
Nice. Setting `observed` now has the desired effect, although the settings from the call to `afex` are not yet honored:

~~~r
afex_aov_mixed <- aov_ez(
    "id"
    , "value"
    , obk.long
    , within = c("hour")
    , between = c("treatment", "gender")
    , observed = "gender"
)

eta_squared(
    afex_aov_mixed
    , generalized = TRUE
)

apa_print(afex_aov_mixed)$table
apa_print(afex_aov_mixed, observed = "gender")$table
apa_print(afex_aov_mixed, estimate = "ges", observed = "gender")$table
~~~

R/apa_print_anova.R
```diff
@@ -61,66 +74,180 @@ apa_print.summary.aov <- function(x, ...) {
...
   afex_aov_intercept <- "(Intercept)" %in% rownames(x$anova_table)
   if(afex_aov_intercept != intercept & afex_aov_intercept) {
     warning("In your call of afex::aov_car() you requested the intercept term, but now you did not (in apa_print 'intercept = FALSE' is the default). Thus, the intercept term will be omitted; make sure this is what you want.")
   }
 
-  if("Anova.mlm" %in% class(x$Anova)) {
+  if(inherits(x$Anova, "Anova.mlm")) {
     summary_x <- summary(x$Anova)
-    apa_print(summary_x, correction = correction, intercept = intercept, ...)
+    apa_print(summary_x, correction = correction, intercept = intercept, .x = x, ...) # apa_print.summary.Anova.mlm
```
crsh, line 248: If we want to carry the `observed` setting from `afex` forward, we could check for the specification of `observed` in the ellipsis and if it is `NULL` default `observed` to this:

~~~r
attr(x$anova_table, "observed") 
~~~

This yields `character(0)` if no observed variables have been specified. Any thoughts?

mariusbarth: I'm already working on this

mariusbarth: Should we really allow the user to specify `observed` differently than what is specified in the model? I'm inclined to overwrite the user's argument (passed to `apa_print()`), or at least warn if there is a disparity

mariusbarth: I just committed a change that will give an informative warning when `observed` and the information from the `afex_aov` object differ.

crsh:
Nice this seems to work as intended:

~~~r
library("papaja")
library("afex")
library("effectsize")

afex_aov_mixed <- aov_ez(
    "id"
    , "value"
    , obk.long
    , within = c("hour")
    , between = c("treatment", "gender")
    , observed = "gender"
)

eta_squared(
    afex_aov_mixed
    , generalized = TRUE
)

apa_print(afex_aov_mixed)$table
apa_print(afex_aov_mixed, observed = "gender")$table
apa_print(afex_aov_mixed, observed = "treatment")$table
~~~

Final ask before I merge: Could we slightly change the wording of the warning to:

~~~
In your call to apa_print(), you specified the model terms `treatment` as observed, whereas in your call to afex::aov_car() you specified `gender` as observed. Make sure that this is what you want.
~~~

mariusbarth:
Done

crsh:
Excellent, thank you! Unfortunately, I get two failing test (`test_custom_effect_sizes.R:32:5`, `test_custom_effect_sizes.R:83:5`). And maybe it would be a good idea to add a test for the newly added observed feature (/runforcover).

mariusbarth:
Hm, I don't have these failing tests. I added a test now

crsh:
Hmm, are you using the latest version of `effectsize`?

mariusbarth:
Yes, I do

mariusbarth (commented):
That makes perfect sense. (Haha, that didn't work as expected)

crsh:
Really weird. I'm running the latest version of all packages, too. Unfortunately, the Travis-CI build does not work at the moment. I'll merge the PR, but we should keep an eye on it.

Thanks again for the excellent work on this PR! 🏅

PR #453, opened 2021-05-05
use centering instead of begin center

mutlusun:
This PR changes the `\begin{center}` to `\centering` in a table environment. The centering environment adds a little space that may add an empty pace in certain circumstances. This is not the case with `\centering`.

See also: https://tex.stackexchange.com/questions/23650/when-should-we-use-begincenter-instead-of-centering


crsh (approved):
That's news to me. Thanks!

Issue #446, opened 2021-04-08
Adding appendix causes special characters to not be displayed correctly


mschrein:
Could be related to #353 and #283 

**Describe the bug**
When adding an appendix in the YAML header, special characters such as umlaute (ö) or the euro sign (€) are no longer displayed correctly. For example, ö is instead displayed as Ã¶. This occurs on the title page, the author note, and the text body. Citations are not affected (neither in-text nor in the reference section).

**To Reproduce**
```
---
title: "title"
shorttitle        : "shorttitle"
author:
  - name          : "Nöl"
    affiliation   : " 1"
affiliation:
  - id            : "1"
    institution   : "University of x"
authornote: |
  Again name Nöl.
abstract: |
   abstract.  
keywords          : ""
wordcount         : ""
bibliography      : ["references.bib"]
floatsintext      : yes
figsintext        : yes
figurelist        : no
tablelist         : no
footnotelist      : no
linenumbers       : yes
mask              : no
draft             : no
csl               : "apa.csl"
documentclass     : "apa7"
classoption       : "man"
output            : papaja::apa6_pdf
appendix          : "appendix.Rmd"
header-includes:
  - \setlength\parindent{24pt}\setlength{\parskip}{0.0pt plus 1.0pt}
  - \raggedbottom
 body-includes:
  - \usepackage{upgreek}
  - \usepackage{multirow}
  - \usepackage{threeparttable}
  - \usepackage{epstopdf}
  - \usepackage{caption}
 ---

Some text body with the € sign.
```

Removing the appendix renders the special characters correctly.

**Additional context**
I already tried several things brought up in other issues (e.g., #350) such as changing the latex enginge to xelatex or manually inserting the special characters via ```'N\"{ö}l'``` (both single and double space) or ```"N`\"{ö}`{=latex}l"``` but nothing worked. All files are UTF-8 encoded and I am using a Windows machine. The addition of appendix in the YAML header does seem to be causal for this problem.

Thanks a lot in advance for your help and your work on this great project.

mariusbarth:
Hi @crsh, I can replicate the error on Windows, too. It seems like the error occurs within `modify_input_file()`, as the temporary .rmd file already contains the encoding-related errors. However, I was hesitant to change the code, because I dimly remember there were some-encoding-related problems with Japanese and Chinese locales.

mariusbarth:
[...]

Issue #353, opened 2020-02-12
Unicode is rendered differently in body vs. in a child vs. in an appendix


kalenkovich:
**Describe the bug**
Unicode (Russian) characters are rendered differently in the body vs. in a child vs. in an appendix. The same string produces three different strings in the `.tex` output:

> Hello, world! ÐŸÑ€Ð¸Ð²ÐµÑ‚, Ð¼Ð¸Ñ€.
> Hello, world! Привет, мир.
> Hello, world! РџСЂРёРІРµС‚, РјРёСЂ.

Only the middle one (from a child document) is correct. Even that one is not rendered in the pdf, but that is a different issue that is certainly not a `papaja` bug. I'll try StackOverflow for that.

**To Reproduce**
Here is a string we will use:

`Hello, world! Привет, мир.`

Create a new document with the `papaja`'s `apa6` template. Replace the "output ..." string in the preamble  with the following:

```yml
output            : 
  papaja::apa6_pdf:
    includes:
      after_body: "appendix.tex"
    latex_engine: xelatex
appendix          : 'appendix.rmd'
```
The `latex_engine: xelatex` setting was suggested by the error message when I tried to knit the document without it. The appendix was added in two ways to see if the results would be different. They were not.

In the body, add the following (replace two inverted commas with three):

```rmd
Hello, world! Привет, мир.

``{r child = "appendix.rmd"}
``

``{r}
papaja::render_appendix('appendix.rmd')
``
```

Add the following string to the `setup` chunk:

```r 
Sys.setlocale(category = 'LC_CTYPE', locale = "Russian")
```

Files `master.rmd` and `appendix.rmd` are attached as an archive.
[papaja_unicode-in-appendix_test.zip](https://github.com/crsh/papaja/files/4190517/papaja_unicode-in-appendix_test.zip)

**Expected behavior**

All four strings (from the body, from the child, from the two identical appendices) are identical, even if not showing correctly.

**Screenshots**
Body and child:

![image](https://user-images.githubusercontent.com/9257050/74311627-f9c7e080-4d80-11ea-8714-28c023a5ad1c.png)

Either of the appendices:

![image](https://user-images.githubusercontent.com/9257050/74313802-6c3abf80-4d85-11ea-89d9-065922eb65ca.png)

**Additional context**
I am on a Windows machine.
I've tried suggestions in #283. I think this is a different issue because of the difference between the body and the child.
[...]

PR #461, opened 2021-07-21
Encoding-related fixes

mariusbarth:
Hi @crsh,

pending feedback from @kalenkovich, I think issues #446 and #353 should be resolved with the changes proposed here. Basically, the code now ensures that whenever a file is read, it is read as unicode (as both Unixes and Windows are capable of). Whenever a file is written, it is written byte-by-byte, because Windows does not seem to support writing unicode, yet. For reference, here is the crucial part of the documentation of `file()`:

> When **writing** to a text connection, the connections code always assumes its input is in native encoding, so e.g. **writeLines has to convert text to native encoding**. writeLines does not do the conversion when useBytes=TRUE (for expert use only), but the connections code still behaves as if the text was in native encoding, **so any attempt to convert encoding** (encoding argument other than "" and "native.enc") **in connections will produce incorrect results.**
>
> When **reading** from a text connection, the connections code, after re-encoding based on the encoding argument, returns text that is assumed to be in native encoding; an encoding mark is only added by functions that read from the connection, so e.g. readLines can be instructed to mark the text as "UTF-8" or "latin1", but readLines does no further conversion. **To allow reading text in "UTF-8" on a system that cannot represent all such characters in native encoding (currently only Windows), a connection can be internally configured to return the read text in UTF-8 even though it is not the native encoding**; currently readLines and scan use this feature when given a connection that is not yet open and, when using the feature, they unconditionally mark the text as "UTF-8". 

This boils down to the following technique:
```
# for reading a new internal function that ensures that the connection is opened correctly:
readLines_utf8()

# for writing
writeLines(text, con, useBytes = TRUE)
```

### DIFF
--- a/R/apa6_formats.R
+++ b/R/apa6_formats.R
@@ -763,14 +763,18 @@ revert_original_input_file <- function(x = 1) {
   return(NULL)
 }
 
+
+
 #' @keywords internal
 
 readLines_utf8 <- function(con) {
   if(is.character(con)) {
     con <- file(con, encoding = "utf8")
     on.exit(close(con))
+  } else if(inherits(con, "connection")) {
+    stop("If you want to use an already existing connection, you should use readLines(), directly.")
   }
   y <- try(readLines(con, encoding = "bytes"))
-  if(inherits(y, "try-error")) stop("File ", summary(con)$description, " seems to be missing.")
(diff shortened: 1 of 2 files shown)


crsh (approved):
Fantastic! Thanks. 🎖️

PR #469, opened 2021-08-04
Hot fix: Safer detection of Chi^2 test in stat_glue(). #154

crsh:
The problem in #154 seems to be that `stat_glue()` tries to determine, whether there is sample size information in an fundamentally unsafe way. This is a hot fix for the issue, but requires a lot more thought.

### DIFF
--- a/R/glue_apa_results.R
+++ b/R/glue_apa_results.R
@@ -223,7 +223,7 @@ stat_glue <- function(x) {
     if(!is.null(x$df.residual)) {
       df <- "(<<df>>, <<df.residual>>)"
     } else if( # Chi^2-Test
-      variable_labels(x$statistic) == "$\\chi^2$" &&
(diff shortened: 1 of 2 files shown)


mariusbarth (approved):
Maybe use `if(identical(variable_label(x$statistic), "$\\chi^2$"))` or `if(any(grepl(variable_label(x$statistic), pattern = "chi^2", fixed = TRUE)))`

mariusbarth:
To further improve stability, I think we could switch to an additional attribute for the `statistic` column (e.g., `attr(x$statistic, "n")`. In the future, we could then drop this attribute if a proper measure of effect size is calculated.

crsh:
Not sure I follow your last comment. Unless I misremember, `n` is used really only when a contigency table test ist reported (as per APA). So I don't think we can drop it with any other effect size measures?

crsh:
I'll merge this PR and open a new issue to improve the handling of `n`.

PR #471, opened 2021-08-05
Adds n as attribute to $statistic for htest

### LINKED ISSUE #470: stat_glue() unsafely searches for sample size information for chi^2-tests
The problem in #154 seems to be that `stat_glue()` tries to determine, whether there is sample size information (i.e. an object named `n`) in an fundamentally unsafe way by searching a parent environment. Unless I misremember, as for now `n` is used only when a contigency table test ist reported (as per APA):

~~~r
smokers  <- c(83, 90, 129, 70)
patients <- c(86, 93, 136, 82)
prop_stat <- prop.test(smokers, patients)
apa_print(prop_stat, n = sum(patients))
~~~

crsh:
This PR addresses #470. I briefly considered passing `n` via `...` to `stat_glue()`, which would grant the flexibility to add all sorts of options if needed. While I think this is generally a useful approach, for the time being there seems to be no need to complicate things

@mariusbarth [suggested](https://github.com/crsh/papaja/pull/469):

> To further improve stability, I think we could switch to an additional attribute for the statistic column (e.g., attr(x$statistic, "n").

This works just as well and has the benefit that this way we provide all information of interest with the `apa_results_table` returned by `apa_print()`. So even if, at a later point, we decide to also use `...` as outlined above, I think adding the attribute makes sense. So for now, we should leave it at that.

### DIFF
--- a/R/apa_print_htest.R
+++ b/R/apa_print_htest.R
@@ -153,6 +153,8 @@ apa_print.htest <- function(
       stop("Please provide the sample size to report.")
     # } else {
     #   n <- paste0(", n = ", n)
+    } else {
+      attr(x$statistic, "n") <- as.integer(n)
     }
   } else {
     n <- NULL
--- a/R/apa_print_htest.R
+++ b/R/apa_print_htest.R
@@ -154,7 +154,7 @@ apa_print.htest <- function(
     # } else {
     #   n <- paste0(", n = ", n)
     } else {
-      attr(x$statistic, "n") <- as.integer(n)
+      attr(x$statistic, "n") <- printnum(as.integer(n))
     }
   } else {
     n <- NULL
--- a/R/glue_apa_results.R
+++ b/R/glue_apa_results.R
@@ -224,8 +224,8 @@ stat_glue <- function(x) {
       df <- "(<<df>>, <<df.residual>>)"
     } else if( # Chi^2-Test
       identical(variable_label(x$statistic), "$\\chi^2$") &&
-      exists("n", where = parent.frame(2), mode = "numeric") &&
-      !is.null(get("n", envir = parent.frame(2), mode = "numeric"))
+      !is.null(attr(x$statistic, "n")) &&
+      is.integer(attr(x$statistic, "n"))
     ) {
       df <- "(<<df>>, n = <<n>>)"
     } else {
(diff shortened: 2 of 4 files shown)


R/apa_print_htest.R
```diff
@@ -153,6 +153,8 @@ apa_print.htest <- function(
       stop("Please provide the sample size to report.")
     # } else {
     #   n <- paste0(", n = ", n)
+    } else {
+      attr(x$statistic, "n") <- as.integer(n)
```
mariusbarth, line 157: Because `x` is the already-typeset `apa_results_table`, I think it would be more consistent to add the *n* as a character via `attr(x$statistic, "n") <- printnum(n)`. This would also ensure that for large *n*, the big mark is properly applied.

R/glue_apa_results.R
```diff
@@ -224,8 +224,8 @@ stat_glue <- function(x) {
       df <- "(<<df>>, <<df.residual>>)"
     } else if( # Chi^2-Test
       identical(variable_label(x$statistic), "$\\chi^2$") &&
-      exists("n", where = parent.frame(2), mode = "numeric") &&
-      !is.null(get("n", envir = parent.frame(2), mode = "numeric"))
+      !is.null(attr(x$statistic, "n")) &&
+      is.integer(attr(x$statistic, "n"))
```
mariusbarth, line 228: With the change above, I think line 228 would become superfluous. Or am I missing something?

tests/testthat/test_apa_print_htest.R
```diff
@@ -270,6 +270,10 @@ test_that(
       )
     )
 
+    expect_identical(
+      attr(two_sample_prop_test_output$table$statistic, "n")
+      , as.integer(sum(patients[3:4]))
```
mariusbarth, line 275: Would have to be `printnum(sum(patients[3:4]))`

mariusbarth (changes requested):
I think we should add the attribute as a character, because (1) the rest of the table is also completely typeset and (2) we still have to apply `printnum()` to it , somewhere

crsh:
Good point, thanks. I implemented all changes as suggested.

mariusbarth (approved):
Perfect :1st_place_medal:

Issue #456, opened 2021-06-07
Incorrect estimate label in `apa_print()` for `emmeans` interaction contrast


shirdekel:
**Describe the bug**
`apa_print()` uses (what seems like to me to be) an incorrect estimate label for `emmeans` interaction contrasts.

**To Reproduce**
``` r
library(papaja)
#> Loading required package: tinylabels
library(emmeans)

model <- lm(mpg ~ vs * am, data = mtcars)

emm <- emmeans(model, ~ vs * am)

emm_contrast_1 <- contrast(emm, "pairwise")

apa_print(emm_contrast_1)
#> $estimate
#> $estimate$`00_10`
#> [1] "$\\Delta M = -5.69$, 95\\% CI $[-10.20,~-1.18]$"
#> 
#> $estimate$`00_01`
#> [1] "$\\Delta M = -4.70$, 95\\% CI $[-9.44,~0.04]$"
#> 
#> $estimate$`00_11`
#> [1] "$\\Delta M = -13.32$, 95\\% CI $[-17.83,~-8.81]$"
#> 
#> $estimate$`10_01`
#> [1] "$\\Delta M = 0.99$, 95\\% CI $[-4.28,~6.27]$"
#> 
#> $estimate$`10_11`
#> [1] "$\\Delta M = -7.63$, 95\\% CI $[-12.70,~-2.56]$"
#> 
#> $estimate$`01_11`
#> [1] "$\\Delta M = -8.62$, 95\\% CI $[-13.89,~-3.35]$"
#> 
#> 
#> $statistic
#> $statistic$`00_10`
#> [1] "$t(28) = -3.45$, $p = .009$"
#> 
#> $statistic$`00_01`
#> [1] "$t(28) = -2.71$, $p = .053$"
#> 
#> $statistic$`00_11`
#> [1] "$t(28) = -8.07$, $p < .001$"
#> 
#> $statistic$`10_01`
#> [1] "$t(28) = 0.51$, $p = .955$"
#> 
#> $statistic$`10_11`
#> [1] "$t(28) = -4.11$, $p = .002$"
#> 
#> $statistic$`01_11`
#> [1] "$t(28) = -4.46$, $p = .001$"
#> 
#> 
#> $full_result
#> $full_result$`00_10`
#> [1] "$\\Delta M = -5.69$, 95\\% CI $[-10.20,~-1.18]$, $t(28) = -3.45$, $p = .009$"
#> 
#> $full_result$`00_01`
[...]

Issue #200, opened 2018-04-04
apa_print for emmeans


andreifoldes:
This is just a feature request, but I think there is great need for apa_print for the emmeans package.

crsh:
I'm currently focusing on implementing support for ANOVA models. As `emmeans` supports *a lot*  of models, support for the output can only be added incrementally. I think the methods now support all follow-up tests for individual cells, pairwise comparisons, and almost all forms of custom contrasts.

I'll try to track what's left to do for ANOVA here:

- [ ] Add support for simple contrasts (see `simple` argument)
- [ ] Add support for `joint_tests()` output
- [ ] Indicate families of tests in result table
- [ ] Indicate which p-values have been adjusted for multiple comparisons

crsh:
I think it would be sensible to use `broom` here but the `emmeans` support is currently somewhat limited. I have submitted a [PR](https://github.com/tidymodels/broom/pull/691) to `broom` and have another one cued. So this issue is deferred until `broom` extends its `emmeans`-support. :/

shirdekel:
What's the update on this? It seems that those PRs have been merged in `broom`.

My specific request (let me know if you'd prefer this in a new issue) is to be able to use `apa_print()` with `emtrends()`, as using it with `emmeans()` currently works fine (except for the CI NAs, I guess):

``` r
  library(papaja)
#> Loading required package: tinylabels
  library(emmeans)

  fiber.lm <- lm(strength ~ diameter * machine, data = fiber)

  emmeans_result <- emmeans(fiber.lm, ~machine)
#> NOTE: Results may be misleading due to involvement in interactions

  apa_print(emmeans_result)
#> $estimate
#> $estimate$A
#> [1] "$M = 40.22$, 95\\% CI NA"
#> 
#> $estimate$B
#> [1] "$M = 41.60$, 95\\% CI NA"
#> 
#> $estimate$C
#> [1] "$M = 38.53$, 95\\% CI NA"
#> 
#> 
#> $statistic
#> $statistic$A
#> [1] "$t(9) = 51.76$, $p < .001$"
#> 
#> $statistic$B
#> [1] "$t(9) = 48.50$, $p < .001$"
#> 
#> $statistic$C
#> [1] "$t(9) = 39.88$, $p < .001$"
#> 
#> 
#> $full_result
#> $full_result$A
#> [1] "$M = 40.22$, 95\\% CI NA, $t(9) = 51.76$, $p < .001$"
#> 
#> $full_result$B
[...]

Issue #95, opened 2016-12-27
Clean up unit tests


crsh:
There is a lot of tedious redundant unit testing happening. Should be fixed before the CRAN release. The `covr` package is helpful to diagnose this.

```{r}
cov <- package_coverage(type = "all", combine_types = FALSE)
shine(cov)
```

mariusbarth:
Just to keep you posted: On my fork, I already streamlined unit testing considerably. Will be happy to merge soon

PR #476, opened 2021-09-02
Extend emmeans-support

crsh:
This PR cleans up the existing `emmeans`-methods for `apa_print()` and extends support. What I have done so far:

- Switch to using `broom::tidy()`and `glue_apa_results()` under the hood
- Clean up unit tests (#95)
- Fix double-mathed infinities (see #456)
- Estimate name guessing for several link functions (see `est_name_from_call()`, #200)
- Support for z-Tests (#200)
- Better formatting of levels of numeric predictors (see `utils.R`)

Some things remain to be done, before this PR can be merged:

- [x] Add indicator for multiplicity correction
- [x] Support `joint_tests()`-output
- [x] Fix failing tests
- [x] Update documention
- [x] Add NEWS entry

I will tackle these next, but I think a review of the changes would be great before I continue. Among other things, it would be great if you could review the choices for the estimate names I have made. Also, maybe there is a way to use `beautify()` here...


R/print_interval.R
```diff
@@ -113,7 +113,7 @@ print_interval.numeric <- function(
       interval <- paste0(
           conf_level
           , "$["
-          , paste(strip_math_tags(x), collapse = ", ")
+          , paste(strip_math_tags(x), collapse = ",~")
```
mariusbarth, line 116: I don't like the idea of using a non-breaking space here. I think a line break between lower and upper boundary of a CI is much better than a line break between, for instance
```
p <
.001
```
Adding a non-breaking space in the CI increases the probability that nasty edge cases like the one I showed here occurs, so I would keep the old behavior. If we would really want to be consistent with that, `p~<~.001` would be of much higher priority than the two values of the CI, I suppose.

R/utils.R
```diff
@@ -342,14 +342,31 @@ beautify <- function(x, standardized = FALSE, use_math = FALSE, ...) {
...
   x <- gsub("\\W", "_", x)                          # Replace non-word characters with "_"
   x
 }
 
 
+#' @rdname sanitize_terms
+#' @method sanitize_terms data.frame
+
+sanitize_terms.data.frame <- function(x, ...) {
+  as.data.frame(lapply(x, sanitize_terms, ...))
```
mariusbarth, line 366: For old versions of R, we should add `stringsAsFactors = FALSE`. Moreover, `check.names = FALSE` would also make sense here (because it is an internal workhorse function and we definitely want to keep character vectors).

R/utils.R
```diff
@@ -376,6 +405,27 @@ prettify_terms <- function(x, standardized = FALSE) {
   x
 }
 
+#' @rdname prettify_terms
+#' @method prettify_terms data.frame
+#' @export
+
+prettify_terms.data.frame <- function(x, ...) {
+  as.data.frame(lapply(x, prettify_terms, ...))
```
mariusbarth, line 413: same as above

R/apa_print_emm_lsm.R
```diff
@@ -52,7 +53,7 @@ apa_print.summary_emm <- function(
   x
   , contrast_names = NULL
   # , stat_name = NULL
-  , est_name = NULL
+  , est_name = "\\widehat{\\theta}"
```
mariusbarth, line 56: I always preferred `\hat` over `\widehat` for `\theta`, think, wide hat is too wide

R/apa_print_emm_lsm.R
```diff
@@ -68,143 +69,186 @@ apa_print.summary_emm <- function(
...
   if(!p_supplied) {
-    warning("Object 'x' does not include test statistics (i.e., p-values).")
+    # warning("Object 'x' does not include test statistics (i.e., p-values).")
 
     df_colname <- NULL
     stat_colnames <- NULL
   } else {
-    dfdigits <- as.numeric(x$df %%1 > 0) * 2
+    df_colname <- names(tidy_x)[grepl("df\\.*", names(tidy_x))]
+    dfdigits <- as.numeric(x[[df_colname]] %%1 > 0) * 2
```
mariusbarth, line 111: This would be a use case for `print_df()`, but I think it will be better to switch to `beautify()`, which in turn applies `print_df()`

R/apa_print_emm_lsm.R
```diff
@@ -68,143 +69,186 @@ apa_print.summary_emm <- function(
...
     dfdigits <- ifelse(is.na(dfdigits), 0, dfdigits) # In case df are Inf
-    multiple_df <- !isTRUE(all.equal(max(x$df), min(x$df)))
-    df_colname <- "df"
-    stat_colnames <- c("statistic", "p.value")
+    multiple_df <- !isTRUE(all.equal(max(x[[df_colname]]), min(x[[df_colname]])))
```
mariusbarth, line 113: `range()`?

R/apa_print_emm_lsm.R
```diff
@@ -68,143 +69,186 @@ apa_print.summary_emm <- function(
...
-    # colnames(contrast_table) <- c("estimate", "conf.int")
-    # contrast_table$ci <- as.character(contrast_table$ci)
-    variable_label(contrast_table) <- c(
-      estimate = paste0("$", est_name, "$")
-      , conf.int = conf_level
+
+    variable_labels(tidy_x[[p_value]]) <- switch(
+      p_value
+      , "p.value" = "$p$"
+      , "adj.p.value" = "$p_{adj}$"
```
mariusbarth, line 200: I just stumbled again over column name `adj.p.value`, which I really despise, but checked the `apa_print.glht()` and saw that we did that atrocity there, too. :D So I will just shut my mouth. Still, I think we should consider changing the label: `p_{adj}` is obviously wrong, it would (at the very least) have to be `p_\\mathit{adj}`. I would even prefer roman letters, and, if this is possible, to add the type of correction, e.g. `p_\\mathrm{BH}`. That would be really helpful, at least for in-text reporting. P.S: I would also love to use upper index and smaller font, such as `p^{\\mathrm{\\smallerBonferroni}`. Consider all of this to be a feature request

R/apa_print_emm_lsm.R
```diff
@@ -218,64 +262,36 @@ apa_print.summary_emm <- function(
...
 
-  if("contrast" %in% factors) {
-    variable_label(contrast_table) <- c(contrast = "Contrast")
+  if(any(tidy_x[, factors] == ".")) {
+    tidy_x[, factors] <- apply(
+      tidy_x[, factors]
+      , 2
+      , gsub
+      , pattern = "\\."
+      , replacement = ""
```
mariusbarth, line 278: Using `fixed = TRUE` would improve performance here

mariusbarth (changes requested):
I added some comments with things that could be changed, but I don't see anything severe (except for the additional arguments to `as.data.frame()` to maintain backwards compatibility with older versions of R). We could indeed use `beautify()`, which would substantially shorten the code. I could add `beautify()` as soon as you are done with this `apa_print()` method. It would be great if you could fix the failing checks before that

R/print_interval.R
crsh: Good point. Maybe we can use `\ ` instead.

R/apa_print_emm_lsm.R
crsh: Ah, I assumed range could not be applied to vectors of lenght 1 but it can. :)

R/apa_print_emm_lsm.R
crsh: Adding the specific correction is already on the list of things to do. I don't really understand why `adj.p.value` is a problem, but either way, we decided to adhere to the `broom`-glossar, which is where this is from.

R/print_interval.R
mariusbarth: Wouldn't that still prioritize keeping the interval together compared with the space between other symbols?

crsh: > Wouldn't that still prioritize keeping the interval together

I'm not really sure, but it should not be a non-breaking space. I would have to try it out. Also, I'm not sure what a "space between other symbols" is, but if you mean `[x.xx, x.xx]`: The space in equation will not be rendered, which is why I considered `~` in the first place.

mariusbarth: What I mean with "space between other symbols" are instances such as the one I described here with the *p* value. If we have a results section with a couple of in-line equations, it is likely that a CI is reported in close proximity to such a p value. If LaTeX would have to decide whether to add a line break within the CI or within the "p = .xxx", it would be forced to add the line break within the "p = .xxx". I think this is undesirable, and we should therefore not add a non-breaking space within the cI.

crsh:
Hi Marius, the most recent commits implement most of your requested changes and add family-wise error correction marks to p values and confidence intervals. Additionally, the families are indicated by letters in the table output. Would appreciate another round of review.

R/apa_print_emm_lsm.R
```diff
@@ -68,143 +74,218 @@ apa_print.summary_emm <- function(
...
-    multiple_df <- !isTRUE(all.equal(max(x$df), min(x$df)))
-    df_colname <- "df"
-    stat_colnames <- c("statistic", "p.value")
+    df_colname <- names(tidy_x)[grepl("df\\.*", names(tidy_x))]
+    multiple_df <- !isTRUE(all.equal(min(tidy_x[[df_colname]]), max(tidy_x[[df_colname]])))
+    p_value <- names(tidy_x)[grepl("p.value", names(tidy_x), fixed = TRUE)]
+    stat_colnames <- c("statistic", df_colname, p_value)
   }
 
+print_df
```
mariusbarth, line 147: extraneous?

tests/testthat/test_apa_print_emm_lsm.R
```diff
@@ -111,26 +66,74 @@ test_that(
...
       tw_me_lsm_output$table[, c("Valence", "estimate", "conf.int")]
       , tw_me_emm_ci_output$table
     )
 
-    expect_error(apa_print(summary(tw_me_lsm, infer = c(FALSE, FALSE))))
+    expect_apa_results(
+      apa_print(summary(tw_me_lsm, infer = c(FALSE, FALSE)))
+      , labels = list(
+        Valence = "Valence"
+        , estimate = "$\\widehat{\\theta}$"
```
mariusbarth, line 90: `\\hat`?

tests/testthat/test_apa_print_emm_lsm.R
```diff
@@ -111,26 +66,74 @@ test_that(
...
+
+    tw_me_emm_output3 <- apa_print(
+      summary(tw_me_emm, infer = TRUE, side = ">")
+    )
 
+    # https://github.com/crsh/papaja/issues/456#issuecomment-901653372
+    expect_apa_term(
+      tw_me_emm_output3
+      , term = "Neg"
+      , estimate = "$\\widehat{\\theta} = 11.00$, 95\\% CI $[7.37, \\infty]$"
```
mariusbarth, line 103: same

mariusbarth (commented):
only some minor things

R/apa_print_emm_lsm.R
crsh: 🙈

mariusbarth (commented):
only minor comments

R/apa_print_emm_lsm.R
```diff
@@ -294,15 +369,81 @@ apa_print.lsmobj <- function(x, ...) {
...
 
-# apa_print_summary_emm_joint_tests <- function(x, ...) {
-#
-# }
+apa_print_summary_emm_joint_tests <- function(x, in_paren, ...) {
 
+  # Retain structuring variables
+  by_vars <- attr(x, "by.vars")
+  if(!is.null(by_vars)) {
+    canonical_table <- canonize(x[, -which(names(x) == by_vars)])
```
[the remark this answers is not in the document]
crsh: Good catch, thanks.

R/apa_print_emm_lsm.R
```diff
@@ -294,15 +369,81 @@ apa_print.lsmobj <- function(x, ...) {
...
+    term_names <- sanitize_terms(unlabel(canonical_table$term))
+  }
+
+  tinylabels::variable_labels(canonical_table) <- c(
+    term = "Effect"
+    , df = "$\\mathit{df}$"
+    , statistic = "$F$"
+    , p.value = "$p$"
+    , df.residual = "$\\mathit{df}_{\\mathrm{res}}$"
+  )
```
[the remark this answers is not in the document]
crsh: Ah yes, but it labels `term` as `Term`, so I will keep the first line to be consistent with the anova-methods.

DESCRIPTION
```diff
@@ -27,6 +27,7 @@ Imports:
   methods,
   rmarkdown (>= 2.4),
   rmdfiltr (>= 0.1.3),
+  stringr,
```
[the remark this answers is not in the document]
crsh: Interesting, I had not seen that conversation. Our dependency graph has three packages that import `stringr`: `knitr` (where attempts are being made to make it a soft dependency or remove it), `rmarkdown` (where, according to my skimming, it could be easily removed), and `broom` (where it's not essential but would take some work to remove). I'll gladly remove the import, but we should consider starting a conversation with the `broom` developers about this as well if we want to avoid the dependency.

crsh:
I think this PR is now ready to be merged.

mariusbarth (approved):

PR #492, opened 2021-11-25
Improves consistency of lookup table for estimates

crsh:
1. Let's consistently use `\\Delta` for estimates of differences instead of these one-off `_d`.
2. Adds lookup information for `p` column to support `prop.test()` with single proportions.

Working on this got me thinking: Maybe we should use something other than `p` to denote proportions to avoid confusions with p-values. How about `\hat\pi`?


mariusbarth:
I think it makes sense to treat the "difference of two means" and the "mean of difference scores" differently. I think this is also what virtually any statistics textbook does.

crsh:
Yes, that makes sense. So we could do `M_D` or `M_\\Delta`? What about `p` vs. something else?

mariusbarth:
> What about `p` vs. something else?

Eid, Gollwitzer, & Schmidt (2017) use `\\pi`, but I think `\\hat\\pi` would be even better.

PR #493, opened 2021-11-25
More robust reimplementation of apa_print()-methods for BayesFactor

crsh:
This is a first stab with extended support for several previously unsupported analyses (e.g. `correlationBF`).

- Object-oriented reimplementation using the new workflow using `canonize()`, `beautify()`, and `glue_apa_results()`.
- Extended unit testing
- Much improved support for diverese outputs, among others`correlationBF()`, `proportionBF()` and `contigencyTableBF()` (including posterior estimates for the former two)
- Removes the `auto_invert` feature because it is fundamentally incompatible with the new workflow (also, it sometimes confused several early users); instead there is now an explicit `reciprocal` argument that applies to all BF alike.
- Several arguments are renamed (some more to come, see below)
- Drops support for `BFBayesFactorList`, which may return at a later point

This is work in progress, but early feedback would be appreciated. Given the extend of the changes the diff may again be "unwieldy", so maybe take it for a spin. :)

## TODO

- [x] In the output rename `rhs` to `model`
- [x] Use `conf.int` (?) instead of `hdi` (cf. `ci`)
- [x] Add option to report `mcmc.error` of Bayes factors
- [x] Add support for interval tests (see open questions)
- [x] Add unit tests for `lmBF()` and `generalTestBF()`
- [x] Update documentation
- [x] Deprecate old arguments

## Open questions

For the interval tests, we do already get an output, but there are two things I'm unsure about. Consider the following example:

~~~r
ttest_onesided <- BayesFactor::ttestBF(
  x = sleep$extra[sleep$group == 1]
  , y = sleep$extra[sleep$group == 2]
  , nullInterval = c(0, Inf)
)
ttest_onesided
~~~
~~~
Bayes factor analysis
--------------
[1] Alt., r=0.707 0<d<Inf    : 0.174135 ±0%
[2] Alt., r=0.707 !(0<d<Inf) : 2.357715 ±0%

Against denominator:
  Null, mu1-mu2 = 0 
---
Bayes factor type: BFindepSample, JZS
~~~

~~~r
apa_print(
  ttest_onesided
  , central_tendency = median
  , iterations = 10000
  , subscript = "+0"
)
~~~
~~~
$estimate
$estimate[[1]]
[1] "$M_d = 0.26$, 95\\% HDI $[0.00, 1.00]$"

$estimate[[2]]
[1] "$M_d = 0.26$, 95\\% HDI $[0.00, 1.00]$"


$statistic
$statistic[[1]]
[1] "$\\mathrm{BF}_{\\textrm{+0}} = 0.17$"

$statistic[[2]]
[1] "$\\mathrm{BF}_{\\textrm{+0}} = 2.36$"


$full_result
$full_result[[1]]
[1] "$M_d = 0.26$, 95\\% HDI $[0.00, 1.00]$, $\\mathrm{BF}_{\\textrm{+0}} = 0.17$"

$full_result[[2]]
[1] "$M_d = 0.26$, 95\\% HDI $[0.00, 1.00]$, $\\mathrm{BF}_{\\textrm{+0}} = 2.36$"


$table
A data.frame with 3 labelled columns:

  estimate     conf.int statistic
1     0.26 [0.00, 1.00]      0.17
2     0.26 [0.00, 1.00]      2.36

estimate : $M_d$ 
conf.int : 95\\% HDI 
statistic: $\\mathrm{BF}_{\\textrm{+0}}$ 
attr(,"class")
[1] "apa_results" "list"
~~~

We always get two Bayes factors---point null vs. interval and point null vs. the inverse interval. The estimates are identical, because they are generated from the unconstrained model.

1. Should we keep the structure as is with the duplicated estimate? It probably doesn't hurt other than making the list more cumbersome to index?
2. What names should we use for the list elements and what should the table look like (it needs something to indicate what the difference is)? The `broom` glossary has several options (`interval`, `alternative`, or `test`; I would probably favor the first).
3. Should we add the prior scale to the table (and strings)? If so, what should that look like (e.g., column name `parameter` or `scale`)?


R/utils.R
```diff
@@ -211,23 +210,32 @@ beautify <- function(x, standardized = FALSE, use_math = FALSE, ...) {
...
+      y[[i]] <- print_num(y[[i]])
     }
   }
 
   # rearrange ----
   y <- sort_columns(y)
 
-  if(!is.null(y$term)) {
-    perm <- term_order(y$term)
+  identifier <- intersect(colnames(y), c("term", "model"))
```
mariusbarth, line 233: `stopifnot(length(identifier) <= 1L)`? I know it should usually not happen that both `term` and `model` are present, but I think such a test will be helpful for developing new methods or refactoring old ones

R/utils_text.R
```diff
@@ -55,6 +55,7 @@ sanitize_terms.character <- function(x, standardized = FALSE, ...) {
   x <- gsub("\\(|\\)|`", "", x)                     # Remove parentheses and backticks
   x <- gsub("\\.0+$", "", x)                        # Remove trailing 0-digits
   x <- gsub(",", "", x)                             # Remove big mark
+  x <- gsub(" \\+ ", "_", x)                             # Remove big mark
```
mariusbarth, line 58: Can you update the comment?

mariusbarth (approved):
nicely done

crsh:
@mariusbarth, do you have any opinions on the open questions? I would say:

1. Yes, keep redundancy (this also avoids having to treat these objects as special cases internally).
2. After some more thought, I think it would be good to have an `alternative` column here that gives the interval.
3. I think it would be nice to somehow report the prior settings alongside the results, but this gets complicated quickly when there are multiple non-tested parameters (think random effects). So for the time being, I would opt to not add any additional table columns for this.

mariusbarth:
Hi @crsh,

I think we can keep the redundancy. If I'm not missing something, the `BFBayesFactor` object that we process here is most frequently not what we would see in a paper---instead, the BF of interest would most frequently be calculated from these two BFs. Or am I missing an important use case, here?

I also like the idea of having a column `alternative`.

Regarding prior specifications, I am also hesitant to add them as an extra column. It is my impression that for now, it is a safer option to let the researcher describe their prior settings.

crsh:
> I think we can keep the redundancy. If I'm not missing something, the BFBayesFactor object that we process here is most frequently not what we would see in a paper---instead, the BF of interest would most frequently be calculated from these two BFs. Or am I missing an important use case, here?

Well, that depends on the question. If, by saying "calculate from these two BFs" you mean deviding on by the other, you get a Bayes factor that compares a positive to a negative effect. Whether this is what you want, really depends. For example, if you are want to test, whether on average participants can detect a signal and you are analysing partcipants d', it doesn't make any sense to specify a hypothesis that posits a negative condition difference; it should be a point null. Hence, the BF in the object is what you want to report. If, however, you are willing to assume that there is definitely an effect but you are not sure which way it goes, than "your" BF would be what you want. I guess it depends on the question and whether you find point null hypotheses useful or not.

> Regarding prior specifications, I am also hesitant to add them as an extra column. It is my impression that for now, it is a safer option to let the researcher describe their prior settings.

So, maybe for the one-sided tests we can have it all. In the column `alternative` we could print something like "Cauchy(r = 0.707)[0, Inf]"?

mariusbarth:
Hm, given your explanation above, it would indeed make sense that a user could select which BF to report. However, given that this is still possible for the reporting strings and the respective row of the results table is easily sub-settable, I would opt against putting too much effort into an additional argument (we could still decide to add such an argument if users want it, later).

Regarding the prior in the results table, would it be self-evident that this is the prior for the effect-size parameter? Or could this result in ambiguity for more-complex models?

mariusbarth:
Is there any way I can help with this one?

crsh:
So I have added support to report MCMC errors and interval hypotheses. The current behavior is as follows:

~~~r
ttest_onesided <- BayesFactor::ttestBF(
    x = sleep$extra[sleep$group == 1]
    , y = sleep$extra[sleep$group == 2]
    , nullInterval = c(0, Inf)
)
apa_print(ttest_onesided)
~~~

~~~
$estimate
$estimate$interval
[1] "$\\Delta M = 0.26$, 95\\% HDI $[0.00, 1.03]$"

$estimate$inverse_interval
[1] "$\\Delta M = 0.26$, 95\\% HDI $[0.00, 1.03]$"


$statistic
$statistic$interval
[1] "$\\mathrm{BF}_{\\textrm{10}} = 0.17$"

$statistic$inverse_interval
[1] "$\\mathrm{BF}_{\\textrm{10}} = 2.36$"


$full_result
$full_result$interval
[1] "$\\Delta M = 0.26$, 95\\% HDI $[0.00, 1.03]$, $\\mathrm{BF}_{\\textrm{10}} = 0.17$"

$full_result$inverse_interval
[1] "$\\Delta M = 0.26$, 95\\% HDI $[0.00, 1.03]$, $\\mathrm{BF}_{\\textrm{10}} = 2.36$"


$table
A data.frame with 5 labelled columns:

  estimate     conf.int        alternative statistic mcmc.error
1     0.26 [0.00, 1.03]  $[0.00, \\infty]$      0.17       0.00
2     0.26 [0.00, 1.03] $[-\\infty, 0.00]$      2.36       0.00

estimate   : $\\Delta M$ 
conf.int   : 95\\% HDI 
alternative: $\\mathcal{H}_1$ 
statistic  : $\\mathrm{BF}_{\\textrm{10}}$ 
mcmc.error : $\\pm\\%$ 
attr(,"class")
[1] "apa_results" "list"       
~~~

So, I have added a column named `alternative` and settled on naming the "terms" `inteval` and `inverse_interval`. In the interesting of finishing up, I have omitted the prior distribution on the tested term for now. We can always add later.

Issue #484, opened 2021-10-07
Appendix heading isn't included in the TOC


shirdekel:
**Describe the bug**
When adding an appendix through the YAML option, the heading isn't included in the TOC.

**To Reproduce**
Knit `index.Rmd` from https://github.com/shirdekel/papaja-reprex-appendix.

**Expected behavior**
The appendix headings to render in the TOC as per their hierarchy.

Knitting the same body content with `bookdown::pdf_document2` and `# (APPENDIX) Appendix {-}` shows the expected behaviour. See `index.Rmd` from https://github.com/shirdekel/bookdown-reprex-appendix.

**Screenshots**
`papaja` TOC in Mac Preview:

<img width="273" alt="Screen Shot 2021-10-07 at 12 44 26 pm" src="https://user-images.githubusercontent.com/49865192/136307611-4a30e549-d715-456a-9a64-3ffc8169f24b.png">

`bookdown` TOC in Mac Preview:

<img width="273" alt="Screen Shot 2021-10-07 at 12 55 12 pm" src="https://user-images.githubusercontent.com/49865192/136308283-52371f46-93e7-4442-be21-15ded9e89e51.png">

**Session info**

```r
xfun::session_info("papaja")                                                                 
R version 4.0.2 (2020-06-22)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Mojave 10.14.6, RStudio 0

Locale: en_GB.UTF-8 / en_GB.UTF-8 / en_GB.UTF-8 / C / en_GB.UTF-8 / en_GB.UTF-8

Package version:
  assertthat_0.2.1  backports_1.2.1   base64enc_0.1.3   bookdown_0.24.1   broom_0.7.9.9000 
  bslib_0.3.0       cli_3.0.1         cpp11_0.4.0       crayon_1.4.1      digest_0.6.28    
  dplyr_1.0.7       ellipsis_0.3.2    evaluate_0.14     fansi_0.5.0       fastmap_1.1.0    
  fs_1.5.0          generics_0.1.0    glue_1.4.2        graphics_4.0.2    grDevices_4.0.2  
  highr_0.9         htmltools_0.5.2   jquerylib_0.1.4   jsonlite_1.7.2    knitr_1.36       
  lifecycle_1.0.1   magrittr_2.0.1    methods_4.0.2     papaja_0.1.0.9997 pillar_1.6.3     
[...]

Issue #249, opened 2018-09-26
Permitting echo = TRUE in appendices


stonegold546:
When I set `echo = TRUE`in an appendix, I get an error when I knit the document.

The alternative I am using is to save the code in a script, read the script and print out the script without echoing the code that does this.

I'm working on a manuscript, and I'd like to provide some syntax in the Appendix.

Thanks.

crsh:
Hi James, sorry for the late reply. Could you provide a minimal reproducible example (e.g., R Markdown files) so I can dig into this?

stonegold546:
Thanks for the reply.

My papaja version is 0.1.0.9842

I create a new papaja markdown file and permit two appendices:

```
---
documentclass     : "apa6"
classoption       : "man"
output            : papaja::apa6_pdf
appendix          :
  - "appendix_a.Rmd"
  - "appendix_b.Rmd"
---
```

Then in appendix_a.Rmd, all I have is the code chunk with:

```
```{r echo = TRUE}
rnorm(50)
```

When I knit, I get the error message:

output file: trial.knit.md

! LaTeX Error: Environment Shaded undefined.

Error: Failed to compile trial.tex. See trial.log for more info.
Execution halted

If I drop `echo = TRUE` from my code chunk, it knits fine.

Is this enough to reproduce the problem? Thanks.

crsh:
Thanks, I'll look into it, but it may take a little while.

crsh:
Wow, I into this for quite a while yesterday and I haven't found a fix for this. Basically, the problem is that the appendix is rendered separately from the main document. When `pandoc` creates the TeX file for the main document it checks whether there are any code chunks to be displayed and if so inserts the definition of some LaTeX environments to display the coded (among them `Shaded` as mentioned in the error message). If the main document does not contain code chunks, the environment definitions are omitted. Now, if the appendix uses `Shaded`, you get the error above.

I haven't found a way to force `pandoc` to include the definitions, so here are some suggestions for quick fixes.

1. The simplest option is to display at least one code chunk (set `echo = TRUE` for at least one chunk) in the main document. This way, the LaTeX environment definitions are added to the document preamble and also work in the appendix.

2. You could also manually include the needed LaTeX environment definitions for the desired highlighting theme in the preamble. This requires that you create a TeX file with the definitions and add it as `header-include`. See [this StackOverflow question](https://stackoverflow.com/a/51649399/914024) for detailed instructions.

stonegold546:
[...]

PR #497, opened 2021-12-15
Switches to bookdown-appendix approach

### LINKED ISSUE #303: Cannot knit manuscript if appendix.rmd is not located in main directory
**Describe the bug**
When knitting a manuscript with an appendix.rmd (specified via yaml header) that is not located in the same directory as the main document, an error is thrown:
```
Quitting from lines 2-3 (main.rmd) 
Quitting from lines NA-4 (main.rmd) 
Fehler in gzfile(file, "rb") : kann Verbindung nicht öffnen
Ruft auf: <Anonymous> ... withCallingHandlers -> withVisible -> eval -> eval -> readRDS -> gzfile
Ausführung angehalten
```

**To Reproduce**
I prepared a minimal reproducible example here: https://drive.google.com/open?id=15XDCE6_SUqFekj-q85_ruGbOB10UaLSd

### LINKED ISSUE #496: Bug in appendix: ampersand appearing instead of 'and'
Hello,

Thank you for such a great package.

I would like to report a bug regarding the appearance, in appendices only, of an ampersand in non-parenthetical citations for references with two authors, where 'and' should appear instead. The error appears in both the apa6 and the apa7 document classes.

It seems as though the filter that distinguishes parenthetical and non-parenthetical citations only works in the main document.

**To reproduce**

1. Create a main document that includes an appendix;
2. Write a *non-parenthetical* citation for a reference with two authors in the main document, as well as in the appendix;
3. Knit the main document (the appendix will be automatically knitted);
4. Notice in the output that the main document has 'and' linking the two authors (correct APA), whereas the appendix has '&'.

Thank you very much for your attention

crsh:
This PR removes my hacky solution to automagically add appendices to the end of the document in favor of the much more elegant bookdown feature (as documented [here](https://bookdown.org/yihui/bookdown/markdown-extensions-by-bookdown.html#special-headers)).

As a consequence it makes the following additional changes:

- `render_appendix()` is now defunct because it is no long needed
- `base64enc` import is removed because it is no longer needed
- Skeletons now use fenced div syntax for bibliography placement

I think this fixes #496 #484 #249 and generally resolves issues like #322 #353

Finally, I imagine this potentially fixes #303. Could you give it a try, @mariusbarth?

This change additionally needs to be documented in the manual.


mariusbarth:
On Ubuntu, this works for
  - word documents
  - pdf documents with class options `man` or `doc`
  
Fails for class option `jou`, works if and only if I write `# (APPENDIX) Appendices {-}`, bot not if I omit the `{-}

crsh:
> Also, for all output formats, works if and only if I write # (APPENDIX) Appendices {-}, but not if I omit the `{-}`

Yes, I think that's expected behavior, but may be worth highlighting in the manual.

crsh:
> Fails for class option `jou`, here is the error message:
> 
> ```
> ! pdfTeX error (ext4): \pdfendlink ended up in different nesting level than \pd
> fstartlink.
> \AtBegShi@Output ...ipout \box \AtBeginShipoutBox 
>                                                   \fi \fi 
> l.655 \end{document}
> ```

Would you be able to provide a minimal reproducible example? The following works for me:

~~~
---
title             : "The title"
classoption       : "jou"

output            : papaja::apa6_pdf
---


# Heading 1
## Subheading 1

# (APPENDIX) Appendix {-} 

# Heading 2
## Subheading 2
~~~

crsh:
@mariusbarth, any idea why the tests are failing? I don't think this is related to a change in our code? Could it be related to a recent update in `tinylabels`?

mariusbarth:
Yes, it is due to some improvements in tinylabels. The tests fail because they depended on unintended side effects. Will fix'em today

crsh:
The LaTeX error seems to be an edge case related to URLs (e.g. in the reference section) breaking across pages. I don't think this is necessarily related to the appendix. Let's make this into a separate issue.

PR #498, opened 2021-12-16
deprecate 'ci', 'conf.level', and 'args_confint' in an orderly manner

mariusbarth:
I think this is the most orderly way to deprecate `ci`, `args_confint`, and `conf.level`. There are rare cases where users could still pass something like `ci` which will then passed to `formatC()` (resulting in an error), but these are instances where the `apa_print()` method never supported any of the above arguemnts or `conf.int`, I suppose.


crsh (approved):
This looks great. Thanks a lot!

PR #501, opened 2021-12-19
Allow gt1 argument for apa_print() and printp()

JeffreyRStevens:
While APA style omits the initial zero for values ranging between -1 to 1, other publishers don't have this requirement, and I publish in non-APA journals and would like to use `apa_print()` for those journals. 

I changed a number of the `apa_print()` functions and `printp()` to allow for initial zeros if the `gt1` argument is included. The default is always to have no inital zeros, but this can be overridden with `gt1 =TRUE`. This involved adding `...` as an argument to some of the functions and checking for the presence of `ellipse$gt1`. 

I tested all of the changed functions with the examples (for functions that had examples), and they all worked fine.

I'm happy to conduct further testing if request, or let me know if there are better ways to implement what I have suggested.

### DIFF
--- a/R/apa_print_emm_lsm.R
+++ b/R/apa_print_emm_lsm.R
@@ -125,7 +125,7 @@ apa_print.summary_emm <- function(
     contrast_table <- rename_column(contrast_table, c("t.ratio", "z.ratio"), "statistic")
     contrast_table$statistic <- printnum(contrast_table$statistic)
     contrast_table$df <- printnum(contrast_table$df, digits = dfdigits)
-    contrast_table$p.value <- printp(contrast_table$p.value)
+    contrast_table$p.value <- printp(contrast_table$p.value, ...)
   }
 
   if(ci_supplied) {
--- a/R/apa_print_glht.R
+++ b/R/apa_print_glht.R
@@ -72,7 +72,7 @@ apa_print.summary.glht <- function(
   ## Format numbers
   contrast_table$estimate <- printnum(contrast_table$estimate, ...)
   contrast_table$statistic <- printnum(contrast_table$statistic, digits = 2)
-  contrast_table$p.value <- printp(contrast_table$p.value)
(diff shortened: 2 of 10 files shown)


crsh:
Hi Jeffrey,

thanks for the proposed changes. First, I think it would be best to propose this change on the `devel` branch because it has considerably evolved since last merged into `master`. Secondly, I wonder whether using the ellipsis for this purpose is the best approach because it does not allow treating formatting of estimates differently from p-values (e.g. number digits). I suspect that a global option specifically for p-values may be better suited (also because you could set it once and then forget it).

All that said, I'm a little torn about this because it is, strictly speaking, beyond the scope of what `papaja` was initially designed to do. I have previously toyed with the idea of abstracting away the particulars of APA formatting defaults and allowing for a separate layer in the programming that specifies formatting defaults (to accomodate various style guides; e.g., leading zeros, number of digits, exact vs. thresholded p-values etc.). This is, however, a *big*  change to the internals, I think. @mariusbarth what do you think?

mariusbarth:
Hi @crsh, I also think that this will be a bigger change that will require some additional thought. However, maybe we could create an additional branch (from the `devel` branch) where we could start work on it. A possible solution might be to create two more wrapper functions `print_estimate()` and `print_statistic()` that we would have to use from within all `apa_print()` methods (which will be easy for all methods that already use `beautify()`).

We could then create a global option such as:

```
options(papaja.number_formatting = list(
  estimate = list(digits = 2L,)
  , statistic = list(digits = 2L)
  , df = list(digits = 2L), # print_df() would only use this if x%%1 != 0, but this could also be controlled via a new argument
  , p.value = list(digits = 3L)
)
```

crsh:
Yes, this is kind of what I had in mind. I think we would want more flexibility by adding customizable glues as well. But this would be a good start. But maybe let's first open a new issue to discuss and discuss some implementation drafts?

PR #509, opened 2022-01-18
Fuse df columns

mariusbarth:
Added new function `fuse_df()` to move degrees of freedom in an `apa_results_table` from column to variable label of `statistic` column.

- I chose the function name `fuse_df()` instead of `fuse_df_columns()` because I think "columns" is a tricky word and the pluralization might also lead to some confusion. However, I don't have a strong opinion on this choice.
- I also updated the glm, glht, and emmeans methods. I could retire apa_glm_res()` along the way. I believe the complete lm/glm code could be refactored easily using the canonize/beautify workflow.
- Some checks are failing (probably because of upstream changes in the effectsize package), but this is probably best addressed on the development or a separate branch.

We could add `fuse_df()` to the examples section of the

- [x] lm/glm,
- [x] glht, and
- [x]  emmeans methods

after deciding on the function name...


R/arrange_regression.R
```diff
@@ -15,6 +15,8 @@ arrange_regression <- function(x, est_name, standardized, conf.int, ...) {
   ellipsis <- list(...)
   summary_x <- summary(x)
 
+  is_glm <- inherits(x, "glm")
+
   # Obtain the name of the test statistic
   if(class(x)[1] == "glm") {
```
crsh, line 21: Should we also use `is_glm` here?

R/arrange_regression.R
```diff
@@ -46,20 +48,19 @@ arrange_regression <- function(x, est_name, standardized, conf.int, ...) {
     , 1
     , function(y) do.call(function(...) print_confint(x = y[utils::tail(names(y), 2)], ...), ellipsis) # Don't add "x% CI" to each line
   )
-  regression_table <- regression_table[, c("term", "estimate", "conf.int", "statistic", "p.value")] # Change order of columns
+  if(!is_glm) regression_table$df <- print_df(x$df.residual) # is NULL for GLMs
```
crsh, line 51: Hmm, that's surprising? Why is that?

R/arrange_regression.R
```diff
@@ -68,8 +69,9 @@ arrange_regression <- function(x, est_name, standardized, conf.int, ...) {
     , statistic = test_statistic
     , p.value   = "$p$"
   )
+  if(!is.null(regression_table$df)) variable_label(regression_table$df) <- "$\\mathit{df}$"
 
-  class(regression_table) <- c("apa_regression_table", class(regression_table))
+  class(regression_table) <- c("apa_results_table", class(regression_table))
```
crsh, line 74: Just out of curiosity, can you remind me, why we (I?) defined a separate class here? 😅

R/fuse_df.R
```diff
@@ -0,0 +1,89 @@
...
+#' @param x    Either the complete output object created by [apa_print()] methods,
+#'   or the `table` component of such objects.
+#' @param quietly Logical. If FALSE (the default), checks if df columns contain non-integer values.
+#' @param ... further arguments passed from an to other methods
+#'
+#' @examples
+#'   apa_out <- apa_print(aov(yield ~ N * P, npk))
+#'
+#'   # Standard output with separate columns for degrees of freedom:
+#'   apa_out
```
crsh, line 17: To highlight what happens here, should we adapt the example to use `apa_out$table` and `fuse_df(apa_out)$table`?

R/fuse_df.R
```diff
@@ -0,0 +1,89 @@
...
+fuse_df.apa_results_table <- function(x, quietly = FALSE, ...) {
+
+  quietly <- isTRUE(quietly)
+
+  if(is.null(x$df) && is.null(x$df.residual)) {
+    if(!quietly) message("There are no df columns to fuse. Returning original input.")
+    return(x)
+  }
+
+  df_columns <- intersect(colnames(x), c("df", "df.residual", "multivariate.df", "multivariate.df.residual"))
```
crsh, line 54: Can we somehow pull this from the names of the lookup table to future-proof this?

R/fuse_df.R
```diff
@@ -0,0 +1,89 @@
...
+
+  for (i in df_columns) {
+    if( length(unique(x[[i]])) > 1L ) {
+      stop("Cannot fuse columns into variable label:\n Degrees of freedom (in column ", encodeString(i, quote = "'"), ") vary by model term.", call. = FALSE)
```
crsh, line 58: Maybe "vary across table rows" because, e.g., emmeans output does not necessarily return a `terms` column.

R/fuse_df.R
```diff
@@ -0,0 +1,89 @@
...
+  x$table <- fuse_df(x$table, quietly = quietly, ...)
+  x
+}
+
+
+
+#' @rdname fuse_df
+#' @export
+
+fuse_df.apa_results_table <- function(x, quietly = FALSE, ...) {
```
crsh, line 45: How about `check_df` for a more descriptive parameter name?

crsh (changes requested):
This is great, thanks! I have added a few minor comments.

My biggest hesitation is the function name. IMHO `fuse_df` is a little opaque, maybe we can find something more descriptive. I don't have a good idea right now, but I'll think some more about this.

I think it's a great idea to use the function in the examples, also in `apa_table()` maybe?

R/arrange_regression.R
mariusbarth: ah, the comment is bs, will remove it

R/fuse_df.R
mariusbarth: :+1:

R/fuse_df.R
mariusbarth: :+1:

mariusbarth:
I just addressed virtually all the changes that you suggested in your review. What is left is 1) deciding on the function name, 2) adding function to the examples sections (where it makes sense), and 3) using `is_glm` in `arrange_regression()`. From reading the code, it was my impression that `arrange_regression()` will become superfluous as soon as we refactor the lm/glm method.

mariusbarth:
`transmute_df()`?

https://www.merriam-webster.com/dictionary/transmute

crsh:
Excellent, thank you!

PR #517, opened 2022-03-18
Attempt at fixing platform-specific number of trailing digits

mariusbarth:
- apa_df() now compares rounded vectors, instead of using the modulus operator
- new implementation should be robust against class extensions (e.g., 'tiny_labelled') of standard vectors
- also updated docs of `apa_df()` to reflect that it is now exported
- as a side effect, integer degrees of freedom are now more frequently reported (for HLM and emmeans methods)

### DIFF
--- a/R/apa_num.R
+++ b/R/apa_num.R
@@ -411,12 +411,14 @@ printp <- apa_p
 print_p <- apa_p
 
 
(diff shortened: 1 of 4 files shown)


crsh (approved):
Great, thanks!

Issue #520, opened 2022-03-23
Anova multiple model comparison 


romeopenheiro:
I am able to do model comparisons for upto 3 models using the `apa_print` function from the papaja package. However, I am interested to do multiple model comparisons (>3) and generating their tables using `apa_table`. Stargazer package does it well. However, the tables there are not apa formatted. It would be nice to be able to have this feature in papaja.

mariusbarth:
Hi @romeopenheiro,

I thought it was possible to report more than three models at once, but now that I tried to draft example code I figured that I get an error when I add more than three models to the list of models. I assume a) that this is a bug but that b) we will be able to fix swiftly.

mariusbarth:
Hi @romeopenheiro, 

We just added a fix that should address your problem. After re-installing **papaja**, it should be possible to compare more than three models.

mariusbarth:
Perfect, thank you for checking!

PR #521, opened 2022-03-29
started refactoring `apa_print.list()`, which solves #520

mariusbarth:
Hi @crsh,

this PR is intended to solve #520, but also to start refactoring of the `apa_print.list()` method.


crsh:
Looks good. Thanks!

Issue #485, opened 2021-10-15
Prune dependency tree


crsh:
Currently, `papaja` has some pretty heavy (indirect) dependecies, which we could strive to remove. Here are some ideas:

- [x] `gsl` keeps giving users and our CI a hard time during installation. We rely on it indirectly through `MBESS` as a suggest. This should be solved shortyl, as `MBESS` has moved `gsl` to suggests.
- [ ] `stringr` imports `stringi`, which takes a long time to install. The following packages we import use `stringr`:
  - `knitr`: The package uses `stringr` quite a bit, but there is an [open PR](https://github.com/yihui/knitr/pull/1552) to remove `stringr` over the medium term.
  - `rmarkdown`: There is only [one call](https://github.com/rstudio/rmarkdown/blob/0af6b3556adf6e393b2da23c66c695724ea7bd2d/R/base64.R#L9) to `stringr`, which can probably be easily removed.
  - `broom`: There are [several uses](https://github.com/tidymodels/broom/search?q=stringr) of `stringr`, but they should, in principle, be easy to remove.
- [ ] `broom`: Has a heavy tidyverse dependency tree. We use it in some `apa_print()`-methods, but there we could switch to our alternative internal workflow (i.e. `canonize()` etc.) or use/support the output of a different package, such as `parameters`. Arguments for `broom` are the large number of supported models and it's widespread use, which makes it penatrable for independent contributors.

PR #522, opened 2022-03-31
Refactor glm methods

mariusbarth:
Hi @crsh,

I refactored the apa_print.lm() method using the canonize-beautify workflow. As of now, `arrange_regression()` and `apa_print.glm()` are not needed, anymore.

I still defined `apa_print.glm <- apa_print.lm` explicitly to maintain backward compatibility with (legacy) code that explicitly calls `apa_print.glm()`.

Moreover, I was able to remove the **broom** dependency in `apa_print.lm()`, so this also partially addresses #485.


crsh (approved):
Looks great! Thanks.

Issue #536, opened 2022-10-18
Incorrect assigment with add_equals = T


andreasnanz:
**Describe the bug**
`apa_p()` assigns incorrectly "=" to values that are < .001 but > .0005.

**To Reproduce**
````
apa_p(0.0006, add_equals = T)
apa_p(0.0005, add_equals = T)
````

**Output**

````
[1] "= .001"
[1] "< .001"
````

**Expected behavior**
I expected that all values that are < .001 get "<" assigned.

**Additional context**
Applies also to `apa_num()` which wraps `apa_p()`.

PR #537, opened 2022-10-27
apa_num: Check if x is too small before rounding.

crsh:
To address #536, I deferred rounding in `apa_num()` until after the checks if the value is small enough to require a prefix (`<` or `>`) have been run.


mariusbarth (approved):
on point

Issue #85, opened 2016-09-16
YAML numbersections: true


lnalborczyk:
Hello,

not really an issue but I'd like to know how to "control" the section-numbering. I've added the following argument `"numbersections: true"` to the YAML header of a "`papaja::apa6_pdf`" Markdown file and surprisingly (or maybe not) section numbering starts at Abstract... 

Thus, Abstract is numbered "1", then the next page containing the title of the file and a wordcount is numbered "2"... and, my first header in text (i.e., "`#Introduction`") appears to be numbered 3...

I would like to start section-numbering at "Introduction" (i.e., my first header in text). In other words, I'd like to "unnumber" Abstract and title.

Is there an easy way to do that ?

Thanks for your help and for the great work you did on this package.

Best regards,

Ladislas

crsh:
Hi, after experimenting a bit with different approaches, the easiest thing I have found is setting `numbersections: false` and adding `\setcounter{secnumdepth}{5}` to the beginning of your document. Alternatively, you can install the current development snapshot (`devtools::install_github("crsh/papaja", ref = "devel")`) in which `numbersections: true` should just do what you expected.

lnalborczyk:
It works perfectly, thanks a lot.

Ladislas

jrennstich:
Unfortunately, when rendering `man` in the current version [0.1.0.994], the numbering starts with authornote (then abstract), with the first headline in body being numbered 3. 

Works like a charm when rendering `doc`.

tfjaeger:
This is an issue again (still?) in R 4.2.1, Papaja 0.0.1. Install the rdevel version does not seem to fix the issue.

(Amazing library, btw, it's been wonderful to have this)

mariusbarth:
Hi @tfjaeger, I just tried to reproduce your issue, but at least on my machine (R 4.2.1 on ubuntu 20.04 with all packages up-to-date) everything works as expected (i.e., section numbering starts with the first header in the main body of the document).

Can you paste your output from `sessionInfo()` here?

GH-Jaeger:
Thanks for the quick reply. To be clear, it works with the `\setcounter{secnumdepth}{5}` workaround. But without it, I get the numbering error. First, here's a minimal version of the doc that causes the issue:

```
---
title             : "title"
shorttitle        : "shorter title"
date              : "`r format(Sys.time(), '%B %d, %Y')`"

author: 
  - name          : "XX"
    affiliation   : "1,2"
[...]

PR #538, opened 2022-10-27
Suppresses numbering of abstract in man-mode.

crsh:
As discussed in #85, when in manuscript mode the abstract is numbered when setting `numbersections: yes`, which is probably undesirable in most cases. This PR should fix this.


mariusbarth:
Hm, "Abstract" is not numbered, but "The title" is. This is probably not completely what you intended, right?

crsh:
Good catch, thanks. I just pushed a commit that should fix this. Would you rereview?

mariusbarth (approved):
I think I cannot do a proper review, but the output with a template document is now exactly what I would expect and desire.

PR #543, opened 2022-11-04
updated github action

### LINKED ISSUE #539: Update R CMD check action
The action is giving the following warning:

> Warning: r-lib/actions/check-r-package@v1 is deprecated. Please update your workflow to use the 'v2' version. Also look at the examples at https://github.com/r-lib/actions/tree/v2/examples because 'v2' workflows are much simpler than 'v1' workflows.

mariusbarth:
- Do not check R3.6, anymore.
- Use V2 of `check-r-package`


crsh (approved):
Looks good to me, thanks!

Issue #559, opened 2023-06-22
Large degrees of freedom displayed with comma


Fritz-theCat:
Hello,
I noticed that degrees of freedom are rendered with commas separating figures of 1000 or more. This is correct behavior for general numbers, but degrees of freedom are specifically excepted from the rule ([https://apastyle.apa.org/instructional-aids/numbers-statistics-guide.pdf](https://apastyle.apa.org/instructional-aids/numbers-statistics-guide.pdf)).
I am unsure if this is new in APA 7 and will be solved in future updates.

```
dat <- data.frame(a = rnorm(20000, 5, 3),
                  b = rnorm(20000, 4, 4))
test <- t.test(dat$a, dat$b)
apa_test <- papaja::apa_print(test)
apa_test$statistic
```

I would be happy if anyone could help me with a workaround.
Best,
Fritz

crsh:
Hi Fritz, thanks for getting in touch. Indeed, this needs to be corrected. Thanks for pointing it out!

PR #565, opened 2023-09-27
Solves issue #559

mariusbarth:
With these changes, degrees of freedom are now reported without a comma as a big mark. However, it is now also possible to specify `big.mark = "anything else"`.


crsh (approved):
Awesome, thanks for this!

PR #566, opened 2023-09-27
Removes outdated test for apa_table(). Fixes #563

### LINKED ISSUE #563: An outdated test for apa_table()
Hi, I just discovered that the following test would fail against the current dev version of **knitr**:
https://github.com/crsh/papaja/blob/1c488f73598ae088a4f0201e767aaa87f28387d3/tests/testthat/test_apa_table.R#L19

The reason that it fails is because of https://github.com/yihui/knitr/commit/547b7cd6b818d06eb9261c8245bb37f7654fc844. The test was working before by chance. I tracked it in Git and found it came from 2015 via @crsh's a2e5521:

https://github.com/crsh/papaja/blob/a2e55211dafea772fc6155798745dfd263842338/inst/tests/testthat/test_apa_table.R#L18

and it was actually for this error:

https://github.com/crsh/papaja/blob/a2e55211dafea772fc6155798745dfd263842338/R/apa_table.R#L100

However, `expect_error()` worked previously only because `added_colnames` was not an argument of `knitr:::kable_latex()`. Now `added_colnames` is accepted in `...`, so `kable_latex()` no longer throws an error, hence the test fails.

Perhaps this test should be removed now? Thanks a lot!

crsh:
As pointed out by @yihui, there we had an outdated test for `apa_table()` that is no longer required because the tested argument has been removed. 

As reported in #563, this test is causing issues for the development version of `knitr`.

This PR removes the test and should resolve the issue in `knitr`. The tests are failing due to an unrelated issue.

### DIFF
--- a/tests/testthat/test_apa_table.R
+++ b/tests/testthat/test_apa_table.R
@@ -16,8 +16,6 @@ test_that(
         , Max = apa_num( max(Recall) )
       )
 
-    expect_error(x <- capture.output(apa_table(descriptives, added_colnames = letters[1:5])))
-
     skip_on_cran()
     rmarkdown::render(
       "test_apa_table.Rmd"


crsh:
@yihui, the update is now available on CRAN.

yihui:
Great! Thank you!

Issue #561, opened 2023-08-01
Running head "DISTORTED PD ESTIMATES IN SEQUENCE LEARNING"


LisaSpitzer:
**Describe the bug**
I am using papaja version 0.1.1.9001, and have followed the online description to use the settings for APA7. When I'm knitting to Word, the running head is correctly displayed on the first page (Running head: my running head), but on the following pages, it is "DISTORTED PD ESTIMATES IN SEQUENCE LEARNING", which I have not defined in any part of my RMD. A quick Google search indicated it might be based on one of your papers. :D

**To Reproduce**
This is my YAML header (I have removed personal info):
`---
title             : "My title"
shorttitle        : "My running head"

author: 
  - name          : "Lisa Spitzer"
    affiliation   : "1"
    corresponding : yes    # Define only one corresponding author
    address       : "Universitaetsring 15, 54296 Trier"
    email         : "my email"
    role: # Contributorship roles (e.g., CRediT, https://credit.niso.org/)
      - "TODO"
 
affiliation:
  - id            : "1"
    institution   : "Leibniz Institute for Psychology (ZPID)"

authornote: |
  Lisa Spitzer, [https://orcid.org/0000-0002-4925-7291](https://orcid.org/0000-0002-4925-7291) 
  
  Enter author note here.

abstract: |
  (removed) 

  
keywords          : "preregistration, open science, meta-research, reproducibility, replicability"
wordcount         : "X"

bibliography      : ["r-references_edited.bib", "rordof.bib"]

floatsintext      : yes
tables            : yes
linenumbers       : yes
draft             : no
mask              : no

figurelist        : no
tablelist         : no
footnotelist      : no

classoption       : "man"
header-includes:
  - |
    \makeatletter
    \renewcommand{\paragraph}{\@startsection{paragraph}{4}{\parindent}%
      {0\baselineskip \@plus 0.2ex \@minus 0.2ex}%
      {-1em}%
      {\normalfont\normalsize\bfseries\typesectitle}}
    
    \renewcommand{\subparagraph}[1]{\@startsection{subparagraph}{5}{1em}%
      {0\baselineskip \@plus 0.2ex \@minus 0.2ex}%
      {-\z@\relax}%
      {\normalfont\normalsize\bfseries\itshape\hspace{\parindent}{#1}\textit{\addperi}}{\relax}}
    \makeatother

csl               : "`r system.file('rmd', 'apa7.csl', package = 'papaja')`"
documentclass     : "apa7"
[...]

PR #567, opened 2023-09-27
Fixes wrong default running head in DOCX on p. > 1

crsh:
In a recent update of the DOCX template, I regenerated this file from our example manuscript. In the process I forgot to set the running head on pages > 2 to `TITLE`, which would be replaced by the correct short title in the postprocessor. Hence the short title of the example manuscript was used in every generated file... yikes
(see  #561)!

This should be fixed by this PR.


mariusbarth (approved):
Blind approval

Issue #562, opened 2023-08-05
Anova doesn't show MSE anymore


Sashpta:
Hello,

**Describe the bug**
Earlier today, when I used apa_table to create a table for an anova it created an anova that had the same rows as the one in the [book](https://frederikaust.com/papaja_man/reporting.html#analysis-of-variance-anova) - $\eta_{G}^{2}, F, df, df_{res}, MSE, p$.
At some point today it swapped $MSE$ for $90% CI$.
This also happens if I use the code from the example in the book...
**To Reproduce**
```r
load("cosmetic_surgery.rdata")
cosmetic_surgery_anova <- afex::aov_ez(
  data = cosmetic_surgery
  , dv = "Post_QoL"
  , id = "ID"
  , between = c("Clinic", "Gender")
)
apa_anova <- apa_print(cosmetic_surgery_anova)
apa_table(
  apa_anova$table
  , caption = "A really beautiful ANOVA table."
  , note = "Note that the column names contain beautiful mathematical copy: This is because the table has variable labels."
)
```
**Expected behavior**
See here https://frederikaust.com/papaja_man/reporting.html#analysis-of-variance-anova (A really beautiful ANOVA table).

**Screenshots**
This is what I get now
![image](https://github.com/crsh/papaja/assets/12276024/7c096cd9-de24-4f01-8d8c-26488ed1ff54)

mariusbarth:
Hi @Sashpta,

this is actually not a bug but a feature, but I admit that it may deserve better documentation. Whether or not MSE is reported in the ANOVA table depends on the estimate that is being calculated (here, typically generalized eta-squared). **papaja** now uses the **effectsize** package to calculate the estimate together with CIs *if it is installed* -- if so, *MSE* is not reported.

**Solution**: If you set use `apa_print(., estimate = "ges")` (the old default), the **effectsize** package is not used and only the points estimate of generalized eta-squared are calculated -- also, *MSE* is reported.

@crsh: If `mse = TRUE` is set, should we always report *MSE* irrespective of what is specified as `estimate`?

Sashpta:
> **Solution**: If you set `apa_print(., estimate = "ges")` (the old default), the **effectsize** package is not used and only the point estimates of generalized eta-squared are calculated -- also, _MSE_ is reported.

That did the trick, thank you! :)

mariusbarth:
[...]

PR #568, opened 2023-09-30
apa_print.<anova>() methods: Report MSEs if requested by user

mariusbarth:
Hi @crsh,

this is my proposal for fixing #562. The global default (`options(papaja.mse)`) is now set depending on whether the **effectsize** package is installed (`FALSE` if installed, `TRUE` if not installed). The rationale is that if **effectsize** is installed, the user will most likely include confidence intervals for their eta-squared. However, I think the reporting strings really did become a bit lengthy if we included both *CIs* and *MSE*s. Therefore, I think reporting either *CI* or *MSE* might be a good compromise -- keeping in mind that this can still be changed by the user.


crsh (approved):
Excellent, thanks. I think this is ready to be merged.

PR #572, opened 2023-10-14
Adding Visual Expectations for Plotting Functions

mariusbarth:
Hi @crsh,

It is still one of my aims to refactor the family of `apa_*plot()` functions in the future. Actually, I already drafted a powerful approach, but the changes are so extensive that it is practically impossible to review these changes all at once and I also want to take some precautions not to break older code.

Therefore, I thought it is wise to first add visual expectations (via the **vdiffr** package) before I start adding theses changes one by one.

This PR is intended to get github actions running, and to gain some experience with **vdiffr**. I will require your review as soon as I have the impression that we can add these tests to the main branch. In the meantime, any feedback on how to improve the testing pipeline is highly appreciated.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -52,7 +52,8 @@ Suggests:
   R.rsp,
   skimr,
   spelling,
-  testthat
+  testthat,
+  vdiffr
 SystemRequirements: Rendering the document template requires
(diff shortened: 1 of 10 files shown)


crsh:
Hah, great. I've been 'considering' this for ages. Thanks for taking the initiative!

PR #578, opened 2024-01-09
fixes #576

### LINKED ISSUE #576: apa_print() cannot parse binom.test() output
**Describe the bug**
The ouput of `binom.test()` can no longer be processed by `apa_print()` because the lookup tables cannot deal with the names pulled from the column `names` attributes.

**To Reproduce**

```r
binom.test(10, 20) |>
  papaja::apa_print()
```
```
$estimate
NULL

$statistic
NULL

$full_result
NULL

$table
A data.frame with 2 labelled columns:

      conf.int p.value
1 [0.27, 0.73]  > .999

conf.int: 95\\% CI 
p.value : $p$ 
attr(,"class")
[1] "apa_results" "list"       
Warnmeldung:
In canonize(y) :
  Some columns could not be renamed: 'number.of.successes', 'number.of.trials', 'probability.of.success'
This implies that your output object was not fully understood by `apa_print()`.
  Therefore, be careful when using its output. Moreover, please visit https://github.com/crsh/papaja/issues and
  file an issue together with the code that generated the output object. In doing so, you help us to fully
  support the type of analysis you just conducted and make papaja a little bit better.
```

**Expected behavior**
The function should return a complete `apa_results`-object.

Any idea what's going on here, @mariusbarth?

mariusbarth:
Hi @crsh,

I think `apa_print()` actually never worked for output from `binom.test()`. This PR adds support for `binom.test()`. I call the estimate `$\\hat\\pi$` (not *p*) to avoid confusion with the *p* value and remove *k* and *n* from the output.

### DIFF
--- a/R/apa_print_htest.R
+++ b/R/apa_print_htest.R
@@ -131,6 +131,9 @@ apa_print.htest <- function(
     x_list
     , stringsAsFactors = FALSE
   )
+  y$number.of.successes <- NULL
+  y$number.of.trials <- NULL
+
   if(!identical(conf_int, list(NULL))) y$conf.int <- conf_int
 
   # sanitize table ----
--- a/R/lookup_tables.R
+++ b/R/lookup_tables.R
@@ -47,6 +47,8 @@ lookup_names <- c(
   , "delta"                     = "estimate"
   , "proportion"                = "estimate"
   , "p"                         = "estimate"
+  # binomial test:
+  , "probability.of.success"    = "estimate"
   # estimate from effectsize package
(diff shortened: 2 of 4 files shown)


crsh (approved):
Awesome, thanks. I thought it maybe this was a side-effect of the changes to our internal workflow or caused by an update to R. Either way, thanks for adding the support! 🏅

crsh:
Could you maybe add a quick line to the NEWS? Other than that this good to merge.

crsh (approved):

PR #580, opened 2024-01-24
Add support for analysis of deviance from car

mariusbarth:
Hi @crsh,

This is my first shot at supporting analysis of deviance output from the **car** package. I need this for one of my manuscripts. There are probably a lot of things we could do to improve the functionality, but this will do in my use case, and I think something is better than nothing.

Best,
Marius


tests/testthat/test_apa_print_anova.R
```diff
@@ -511,5 +511,110 @@ test_that(
...
+    )
+
+    car_lr_out   <- apa_print(car::Anova(glm.D93, type = 3, test.statistic = "LR"))
+    car_wald_out <- apa_print(car::Anova(glm.D93, type = 3, test.statistic = "Wald"))
+    car_f_out    <- apa_print(car::Anova(glm.D93, type = 3, test.statistic = "F"))
+
+    expect_identical(
+      car_lr_out$full_result
+      , list(
+        outcome = "$\\chi^2(2) = 5.45$, $p = .065$"
```
crsh, line 596: Would it make sense to expect that these are identical to the LRT results from above?

R/apa_print_anova.R
```diff
@@ -415,6 +416,30 @@ apa_print.anova <- function(
...
     )
+  # stats::anova.glm() and car::Anova.glm
+  } else if(any(grepl("Deviance", object_heading))) {
+    x$Term <-rownames(x)
+    y <- canonize(x)
+    y <- remove_residuals_row(y)
+    if(all(colnames(x) != "F values")) y$df.residual <- NULL
+    if(any(colnames(x) == "Cp")) y$df <- NULL
+
+    if(is.null(y$statistic)) {
```
crsh, line 427: Hmm, this I don't understand. Shouldn't `canonize` do this?

crsh (commented):
Hi Marius, it has take me a long time, sorry about that. I think this looks good, I just had two minor questions. Maybe you can help me understand the code a little better. Cheers!

R/apa_print_anova.R
mariusbarth: I would consider these fixes specific to analysis of deviance that do not fit into a generic function such as `canonize()` -- moving this to `canonize()` would imply that we would need a list of test statistics that should be reported with degrees of freedom and a list of test statistics that should not be reported with degrees of freedom.

tests/testthat/test_apa_print_anova.R
mariusbarth: I would opt against that -- in this case both are identical, but in principle the results from a type-1 and a type-2 analysis could differ.

crsh: Ah, of course. Sorry.

R/apa_print_anova.R
crsh: I'm not sure I understand. This section does not deal with degrees of freedom, as far as I can tell. I'm referring here to the section where `statistic` is null and is replaced by `deviance`.

mariusbarth: I accidentally deleted my last comment.

Maybe the most concise answer is the following: In one very specific case (chi-squared tests for generalized linear models), a column named `Deviance` contains the *difference in deviance* which is distributed as chi-squared, and is used as the test statistic.

In `canonize()`, however, we rename `Deviance` to `deviance`, not `statistic`, because columns of this name typically do not contain the test statistic, but simply the deviance of the model (in addition to the test statistic contained in another column). Because column `Deviance` containing the test statistic is the exception, not the rule, I would opt for sticking with the current behavior of `canonize()` and deal with the exception within the specific method.

crsh: What a mess. That makes sense, thanks for clarifying!

Issue #583, opened 2024-02-14
Inconsistencies in apa_print.emmGrid()


mariusbarth:
`apa_print.emmGrid()`returns an output object that does not fully comply with the structure of what `apa_print()` should return.

To illustrate:
```
$estimate
$estimate$Regularfasterthannonregular_mixeddeterministic
[1] "$\\Delta M = -66.24$, 95\\% CI $[-75.43, -57.05]$"

$estimate$Regularfasterthannonregular_probabilistic
[1] "$\\Delta M = -13.96$, 95\\% CI $[-22.47, -5.45]$"

$estimate$Mixeddeterministicfasterthanprobabilistic_regular
[1] "$\\Delta M = -47.55$, 95\\% CI $[-62.10, -33.01]$"

$estimate$Mixeddeterministicfasterthanprobabilistic_nonregular
[1] "$\\Delta M = 4.73$, 95\\% CI $[-6.92, 16.37]$"


$statistic
$statistic$Regularfasterthannonregular_mixeddeterministic
[1] "$t(98) = -14.30$, $p < .001$"

$statistic$Regularfasterthannonregular_probabilistic
[1] "$t(98) = -3.25$, $p = .002$"

$statistic$Mixeddeterministicfasterthanprobabilistic_regular
[1] "$t(98) = -6.49$, $p < .001$"

$statistic$Mixeddeterministicfasterthanprobabilistic_nonregular
[1] "$t(98) = 0.81$, $p = .423$"


$full_result
$full_result$Regularfasterthannonregular_mixeddeterministic
[1] "$\\Delta M = -66.24$, 95\\% CI $[-75.43, -57.05]$, $t(98) = -14.30$, $p < .001$"

$full_result$Regularfasterthannonregular_probabilistic
[1] "$\\Delta M = -13.96$, 95\\% CI $[-22.47, -5.45]$, $t(98) = -3.25$, $p = .002$"

$full_result$Mixeddeterministicfasterthanprobabilistic_regular
[1] "$\\Delta M = -47.55$, 95\\% CI $[-62.10, -33.01]$, $t(98) = -6.49$, $p < .001$"

$full_result$Mixeddeterministicfasterthanprobabilistic_nonregular
[1] "$\\Delta M = 4.73$, 95\\% CI $[-6.92, 16.37]$, $t(98) = 0.81$, $p = .423$"


$table
A data.frame with 6 labelled columns:

                                                                                                       contrast estimate         conf.int statistic
[...]

PR #584, opened 2024-02-16
Issue583

mariusbarth:
Hi @crsh,

so this is an attempt at fixing what has been described in issue #583.

- I added a test for row names in `expect_apa_results()` to protect against future glitches.
- Row names are not added to the results table, anymore.
- The term names are now consistently constructed via `sanitize_terms()`.
- `sanitize_terms()` now always replaces multiple consecutive underscores with a single underscore.

In addition to what I implemented here, we could consider `sanitize_terms()` to always remove leading underscores.


crsh (approved):
Awesome, thanks!

crsh:
Regarding your suggestion on `sanitize_terms()`, I think this would make sense. If you want to add this to this PR, feel free and also feel free to merge.

PR #590, opened 2024-08-01
Violin plots

mariusbarth:
Hi @crsh,

I was finally able to implement violin plots, with new function `apa_violinplot()`. Example:

```
data("obk.long", package = "afex")

apa_violinplot(
  data = obk.long
  , dv = "value"
  , id = "id"
  , factors = c("treatment", "phase")
  , args_points = list(bg = 3:5)
  , ylim = c(0, 12)
)
```

grafik

The calculation of densities may be customized via `args_density`, the appearance of violins may be customized via `args_violins`. Hope you like this proposal.


crsh (approved):
Nice work. I think this is good to merge.

Issue #608, opened 2025-10-31
write_bib() fails for packages with more than one citation of the same bibtype


tfjaeger:
In R 4.5.1, papaja version 0.1.4, create_bib fails for libraries like `brms`, which lists two citations. 

This failure seems to be caused by this code chunk:

```
if (length(cite) > 1) {
            bibtypes <- unlist(cite$bibtype)
            pref_entry <- type_pref[tolower(type_pref) %in% tolower(bibtypes)][1]
            cite <- if (is.na(pref_entry)) 
                cite[[1]]
            else cite[[bibtypes == pref_entry]]
        }
```

which throws an error "Error in s[[i]] : attempt to select more than one element in vectorIndex". Changing `cite[[bibtypes == pref_entry]]` to `cite[[which(bibtypes == pref_entry)]]` removes that error and leads to the desired behavior.

crsh:
Resolved by #610.

tfjaeger:
[celebrate]     Jaeger, Florian reacted to your message:
________________________________
From: Frederik Aust ***@***.***>
Sent: Tuesday, January 13, 2026 3:56:59 PM
To: crsh/papaja ***@***.***>
Cc: Jaeger, Florian ***@***.***>; Author ***@***.***>
Subject: [EXT] Re: [crsh/papaja] write_bib() fails for packages with more than one citation of the same bibtype (Issue #608)


Closed #608<https://urldefense.com/v3/__https://github.com/crsh/papaja/issues/608__;!!CGUSO5OYRnA7CQ!a0YZJStAFSitdSXFL9GJFdaAbb6EUKZ54akwNxuO8sZEswWJQ1FatN26sv8mu-FdBC3VAADB_KzJsRvJpwbhcI8lcdvjcQ$> as completed.

PR #610, opened 2025-12-06
Fixing write_bib for packages with multiple citation entries

tfjaeger:
Addressing [issue 608](https://github.com/crsh/papaja/issues/608).

Currently, papaja::r_ref or create_bib fails for me for any package with more than one citation entry. This should address that issue. Tested locally.


tfjaeger:
Please let me know if you would like me to add the PR to some other branch instead. Thank you for maintaining papaja!

PR #619, opened 2026-05-17
Fix indexing in citation selection logic

tfjaeger:
The previous code still could not handle packages with multiple citations of the same bib_type. When this situation occurs (as it does, e.g., for the package brms 2.23.0, which lists three citations of type "Article"), the new approach selects the first of those entries.

Alternatively, one could change this part to allow multiple citation entries (the subsequent code seems to be set up with this in mind), but it is unclear why one would check for a preferred bib_type if one is willing to accept multiple bib entries per package anyway. 

Additionally, constraining the citations to one per package makes sense in order to avoid citation inflation.


R/r_refs.R
```diff
@@ -124,6 +125,14 @@ create_bib <- function(x, file, append = TRUE, prefix = "R-", type_pref = c("Art
         }
         entry[[ent]]
       })
+      if (length(entry) > 1) {
+          warning(
```
crsh: I think it would be sufficient to make this a `message()`.

crsh (commented):
I've tried to test this, but I get various errors. Could you try and create some unit tests for this in `tests/testthat/test_r_refs_cite_r.r`? Maybe the following is a good jumping off point:

```r
context("r_refs() and cite_r()")

test_that(
  "r_refs() writes out all references"
  , {
    skip_on_cran()
    
    library("papaja")
    library("brms")
    r_refs("./test_cite_r_all.bib", type_pref = "not-available", tweak = TRUE)
    r_refs("./test_cite_r_preferred.bib", type_pref = "Article", tweak = TRUE)
  }
)
```

R/r_refs.R
tfjaeger: Changed.

tests/testthat/test_r_refs_cite_r.R
```diff
@@ -0,0 +1,15 @@
...
+
+test_that(
+  "r_refs() writes out all references"
+  , {
+    skip_on_cran()
+
+    library("papaja")
+    library("rmarkdown")
+
+    r_refs("./test_cite_r_all.bib", type_pref = "not-available", tweak = TRUE, append = FALSE)
```
crsh: Could you use temporary files for this? For example:

```r
cite_r_all  <- tempfile(fileext = ".bib")
r_refs(cite_r_all, type_pref = "not-available", tweak = TRUE, append = FALSE)
```

R/r_refs.R
```diff
@@ -31,10 +31,11 @@
 #'   If a package provides citation information in a `CITATION` file, a
 #'   reference is selected based on the preferred order of reference types
 #'   specified in `type_pref`. By default, available articles are cited rather
-#'   than books. If no reference of the specified types is available, the first
-#'   reference is used. If multiple references of the preferred type are given
-#'   all of them are cited. Finally, if no `CITATION` file exists a reference
-#'   is generated from the `DESCRIPTION` file by \code{\link[utils]{citation}}.
+#'   than books. If no reference of the specified types is available, or if
+#'   multiple references of the preferred type are given all of them are cited
+#'   (and a warning is generated). Finally, if no `CITATION` file exists a
```
crsh: Technically, a message is produced. I know this is nitpicking, but while you are addressing the comment on the test, maybe you can fix this as well.

crsh (changes requested):
Thank you @tfjaeger, this is basically good to merge. Just one more minor comment and a nitpicking issue that I raise only because there is one other thing to address. Sorry for the lengthy process and the many loops.

R/r_refs.R
tfjaeger: No worries. Done!

tests/testthat/test_r_refs_cite_r.R
tfjaeger: Done

crsh (approved):
Thanks! Good to go.

PR #620, opened 2026-05-28
Fix apa_num.integer() to honor big.mark formatting

wali-reheman:
When numerals = TRUE (the default), apa_num.integer() returned as.character(x) immediately and never checked for formatting arguments like big.mark. So apa_num(7000L, big.mark = ",") returned "7000" instead of "7,000". Now if any formatC arguments are present in ..., the function routes to formatC instead of short-circuiting.


mariusbarth (changes requested):
Hi @wali-reheman, 

Thank you for spotting this unexpected behavior. I agree that the `formatC()`'s formatting arguments should be respected. However, I would opt for the simpler solution to not whitelist some arguments but to simply call

```r
if(numerals) return(formatC(x, ...))
```
This way, partial matching of `formatC()`'s arguments will be preserved, and the code is simpler. @crsh, any thoughts on this?

PR #622, opened 2026-08-11
Add apa_print.glmmTMB() method

mariusbarth:
Hi @crsh,

This PR adds `apa_print()` support for models estimated with **glmmTMB**.

To avoid duplicate code, I chose the following approach: `apa_print.glmmTMB()` is an alias of `apa_print.merMod()`, which is surgically extended to also handle these new model objects.

Best,
Marius


R/apa_print_merMod.R
```diff
@@ -93,24 +104,21 @@ apa_print.merMod <- function(
 
 
   # Rearrange ----
-  x_summary <- summary(x)
-
-  res_table <- as.data.frame(
-    x_summary$coefficients
-    , row.names = NULL
```
crsh, line 100: Does this change still remove rownames for the results table when merMod-objects are passed?

tests/testthat/test_apa_print_merMod.R
```diff
@@ -44,6 +46,16 @@ test_that(
         , p.value   = "$p$"
       )
     )
+    expect_apa_results(
```
crsh, line 49: Does it make sense to also check the reportable strings and values?

crsh (commented):
Awesome! Left two minor comments.

R/apa_print_merMod.R
mariusbarth: `row.names = NULL` actually does ***not** suppress* row names -- this is a common misconception that seems to shine through this line of code that was probably written by myself. ;) 

That being said, the following line reliably accomplishes what we want:
```
rownames(res_table) <- NULL
```

tests/testthat/test_apa_print_merMod.R
mariusbarth: It does. I just added one test that I would consider most diagnostic.

crsh (approved):
Nice, thank you!

crsh:
If you like, you can also reknit the README to show the newly supported methods. But feel free to just merge the PR. The README will be reknit eventually anyway.

